PR #243, opened 2023-05-27
Add a converter for QASM 2.0 to 3.0

### LINKED ISSUE #223: OpenQASM 2.0 --> OpenQASM 3.0 conversion function
Write a function in `qbraid.interface.qbraid_qasm.tools.py` that converts an OpenQASM 2.0 program / string to an OpenQASM 3.0 program / string.

Resources:
- https://openqasm.com/
-  https://github.com/openqasm/openqasm
- https://github.com/openqasm/openqasm/tree/OpenQASM2.x

Example usage:

```python
>>> print(qasm_str)
OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
creg c[2];
h q[0];
cx q[0],q[1];
measure q[0] -> c[0];
measure q[1] -> c[1];
>>> print(convert_to_qasm3(qasm_str))
OPENQASM 3.0;
bit[2] b;
qubit[2] q;
h q[0];
cnot q[0], q[1];
b[0] = measure q[0];
b[1] = measure q[1];
```

One rough example of a way you could test for coverage:

```python
from qbraid import circuit_wrapper
from qbraid.interface import random_circuit
from qbraid.interface.qbraid_qasm.tools import convert_to_qasm3

from braket.circuits.serialization import IRType

for _ in range(100):
    circuit = random_circuit("braket")
    qasm2_str = circuit_wrapper(circuit).transpile("qasm")
    qasm3_test = convert_to_qasm3(qasm2_str)
    qasm3_expected = circuit.to_ir(IRType.OPENQASM).source
    assert qasm3_test.strip() == qasm3_expected.strip()
```

TheGupta2012:
> tags : `unitaryhack` `interface` `compatibility`

## Changes:
Adds a new function `convert_to_qasm_3` in `qbraid/interface/qbraid_qasm/tools.py`

## Details or comments:
Fixes #223 
According to the [OPENQASM](https://openqasm.com/) specification, many constructs of v2.0 are compatible with v3.0. The changes are as follows - 

1. Register declarations have been changed from `qreg q[1];` to `qubit[1] q;` and `creg c[1];` to `bit[1] c;`
2. Opaque operations are no longer parsed in v3.0 
3. Measure operations are changed from `measure q[i] -> c[j];` to `c[j] = measure q[i];`. This also supports multi-qubit measurement (given the register sizes are the same) 
4. Besides these, the `CX` gate is no longer a primitive operation. It is implemented using the `ctrl @` construct and its definition is readily supplied in the standard header `stdgates.inc`. This is equivalent to the `qelib1.inc` header present in v2.0

Some helper functions are used in code to change different constructs from v2.0 to v3.0. I have also included a semantic check using `qiskit` before converting to v3.0.

One concern I have is about the *keywords of v3.0 being used as identifiers in semantically valid v2.0 programs*. Does that check need to be incorporated at this level? 

Moreover, unit tests for each construct are added

## Checklist before merge
- [ ] More unit tests?
- [ ] Misc
### General 


TheGupta2012:
I am not sure about the test fail as the unit test failure was present in files unaffected by this commit.

ryanhill1:
> I am not sure about the test fail as the unit test failure was present in files unaffected by this commit.

@TheGupta2012 Yes, seems like a token is missing which would just be a GitHub actions permissions error. I'll take a look and lyk what I find! Should be an easy fix

ryanhill1:
@TheGupta2012 Ok, try syncing with latest commit to main and see if that fixes it. If not can try one other approach

TheGupta2012:
> @TheGupta2012 Ok, try syncing with latest commit to main and see if that fixes it. If not can try one other approach

Hey, thanks @ryanhill1! Seems like that was indeed the issue.

ryanhill1 (changes requested):
Your progress looks great so far! It seems like your code covers all of the conversions for gates from [`stdgates.inc`](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/qasm/libs/stdgates.inc). As the next step, we'd also like to support conversions of gates from [`qelib1.inc`](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/qasm/libs/qelib1.inc). Unlike in qasm2, you won't just be able to declare these gates with an "includes" statement. In qasm3, you will have to dynamically create a "custom" gate that implements each one, based on it's definition. Instead of hard-coding each case, you can maybe try to incorporate the above linked file, and read-in the gate definitions as needed as you parse through and convert the qasm2 string. Or, you take your own approach!

Below is some code that you can use to test your solution. To run it, you will have to first install the experimental qiskit qasm3 library:

```bash
pip install 'qiskit-terra[qasm3-import]'
```

```python
from qiskit.qasm3 import loads

from qbraid.interface import circuits_allclose, random_circuit
from qbraid.interface.qbraid_qasm.tools import convert_to_qasm_3

for _ in range(100):
    circuit = random_circuit("qiskit")
    qasm2_str = circuit.qasm()
    qasm3_str = convert_to_qasm_3(qasm2_str)
    circuit_test = loads(qasm3_str)
    assert circuits_allclose(circuit, circuit_test)
```

ryanhill1:
@TheGupta2012 Awesome, these updates are looking great so far! It seems like the qiskit team has already gotten back to you on the issue you raised: https://github.com/Qiskit/qiskit-qasm3-import/issues/11#issuecomment-1568505732. If they're slow pushing the changes on their end, maybe you can separate out the unit tests for the effected gates, and add a `@pytest.skip` decorator so that we still have the tests for later, but for now know that they require revisions from qiskit to run.

In terms of other random tests error, if they continue to impact your progress, let me know and I can try to help debug.

TheGupta2012:
Hi @ryanhill1, I discovered one more bug in terra :smiley_cat:. Its related to the `u0` gate conversion.

Besides that, I have added the unit tests for custom gates and the decorated tests pertaining to the active bugs. 

An average of 80% of random tests are passing with some errors in qasm2 parsing. 

I saw that although the circuit structures (of the two circuits being compared) looked the same, the names of the gates may be the culprit for circuit inequality. Do you think converting circuits to unitary matrices and then comparing them would help? 

If you had time, could you also look at the failures? This was a script that I was using for the same - 

```python3 

import os

import pytest
from qiskit.qasm3 import loads
from qbraid.interface.qbraid_qasm.tools import convert_to_qasm_3

correct = 0
excepts = 0
i = 0
for _ in range(100):
    try:
        circuit = random_circuit("qiskit")
        qasm2_str = circuit.qasm()
        qasm3_str = convert_to_qasm_3(qasm2_str)
        circuit_test = loads(qasm3_str)
        eq = circuits_allclose(circuit, circuit_test)
        if eq:
            correct += 1
        else:
            print(f"Failed test {i}")
            print(circuit.draw())
            print(circuit_test.draw())
            i += 1
    except Exception as e:
        print(e)
        print("Qasm 2 string : ")
        for line in qasm2_str.splitlines():
            print(line)
        print()
        excepts += 1
print("Exceptions :", excepts)
print("Correct circuits : ", correct)
print("Incorrect : ", 100 - correct)
```

TheGupta2012:
Thanks for the suggestion @ryanhill1 ! I see ~ 95% of tests pass when testing for single qubit gates and depth = 1. 
All the failed tests correspond to the `u` gate, as well as any gate utilizing its definition. The source for this may indeed be the global phase as the native `U` gate for qasm3 differs by a global phase from the qasm2 implementation.

I have raised [an issue](https://github.com/Qiskit/qiskit-qasm3-import/issues/12) regarding this problem on the repository and I hope it gets resolved soon.

ryanhill1:
@TheGupta2012 Ok, that sounds good. Same as before, if they're slow to make the change, then we can just note the inconsistency and skip the impacted test cases.

TheGupta2012:
Hi @ryanhill1 , the randomized test is now passing after excluding the problematic gates which we were discussing above. I have also added a note in the docstring of the test to reference the discussion about how the problematic gate list was generated.

Let me know if any changes are required!

qbraid/interface/qbraid_qasm/test/test_qasm_utils.py
```diff
@@ -28,3 +40,227 @@ def test_qasm_depth():
...
+    rccx q[0], q[1], q[2];
+    rc3x q[0], q[1], q[2], q[3];
+    c3x q[0], q[1], q[2], q[3];
+    c3sqrtx q[0], q[1], q[2], q[3];
+    c4x q[0], q[1], q[2], q[3], q[4];
+    """
+    _check_output(convert_to_qasm_3(test_unsupported), test_unsupported_expected)
+
+
+def test_random_conversion_to_qasm3():
```
ryanhill1, line 169: For this, when we generate 200 random circuits and throw out ones with unsupported gates in a single test function, we don't actually know how many tests we're running, so going forward if something breaks, it's hard to figure out where the error is or how badly things are breaking.

Instead, let's create a `@pytest.fixture` or equivalent helper function that is called beforehand to generate a list of supported test circuits. In this function, you can use your same code to generate random circuits, check if each contains any unsupported gates, and if not, append to a testing list until it reaches length 50 or 100.

Then, you can pass the list of supported qasm strings / circuits to this test function using `@pytest.mark.parametrize` and run that number of clean, isolated tests 🚀

qbraid/interface/qbraid_qasm/tools.py
```diff
@@ -54,3 +56,99 @@ def _convert_to_contiguous_qasm(qasmstr: str, rev_qubits=False) -> QASMType:
...
+
+def convert_to_qasm_3(qasm_2_str: str):
+    """Convert a QASM 2.0 string to QASM 3.0 string
+
+    Args:
+        qasm_2_str (str): QASM 2.0 string
+    """
+    try:
+        # use inbuilt method to check validity
+        _ = QuantumCircuit.from_qasm_str(qasm_2_str)
```
ryanhill1, line 133: If we've implemented our conversion function correctly, we shouldn't need qiskit's help identifying invalid qasm strings. Instead, our function should have it's own error handling to identify and raise exceptions when things aren't right! As a bonus for catching our own errors, it's one less import at the top, and we cut down the runtime by 1/2 (one pass over the circuit instead of two, which doesn't make a difference for small circuits, but is good practice especially if used in other functions that perform subsequent conversions)

ryanhill1 (changes requested):
These changes look great! Just two final very small, nit-picky updates, and then everything should be good to go. Thanks for your hard work and diligence on this issue  🎉

qbraid/interface/qbraid_qasm/tools.py
TheGupta2012: Yes, that does seem a good approach @ryanhill1 but then the current line-by-line parsing method would not work. If we were to catch _arbitrary errors_ in qasm2 strings, a parser would be required for the conversion which can report those.

I saw that there's already one built as the `QasmParser` class in `qbraid/transpiler/cirq_qasm/qasm_parser.py` which could be tweaked to build a qasm3 string on-the-fly but I wanted to first confirm if that would indeed be required.

It is also highly probable that I'm missing something and the parser is not required after all!

qbraid/interface/qbraid_qasm/test/test_qasm_utils.py
TheGupta2012: [9af229b](https://github.com/qBraid/qBraid/pull/243/commits/9af229b56139c9bd6fc4e475e69ec6f26a7c7363) should add this

qbraid/interface/qbraid_qasm/tools.py
ryanhill1: That's a fair point. Since we're more swapping operators than parsing, I guess the error handling would be less straight forward. So for now this approach is ok. But going forward we'll likely integrate it in with our internal `QasmParser`, as we don't want to be dependent on qiskit to determine what can or can't be converted.

As one simple example, let's say for whatever reason qiskit's  `from_qasm_str` method is updated and has a bug. Then, it's possible that even for a valid qasm string, the qiskit check could fail and raise an exception, when the conversion would have worked on our end. But like I mentioned, we can address this later on when we integrate it. Thanks for your thoughts!

TheGupta2012:
Hi @ryanhill1 , so sorry that I forgot to update the `requirements.txt` file with the `qiskit_qasm3_import`. Should I open another PR for that or you'll bump the requirements?

ryanhill1:
@TheGupta2012 No problem, small fix I got it!

PR #347, opened 2023-08-13
Format counts now by default plots only non-zero values (can change this with removeZeros parameter)

pranavkakhandiki:
*issue tag, if exist*

## Checklist before merge

### General 


ryanhill1 (changes requested):
This looks great! Just two quick changes before merging:

1. We want to keep the original test case for when `removeZeros = False`. So instead of modifying the existing test case, might be better to write a new separate one that tests new default `removeZeros = True`. Let me know if that makes sense.

2. Fix trailing white space from formatting check failed actions: https://github.com/qBraid/qBraid/actions/runs/5845466321/job/15876716285#step:5:17

qbraid/devices/tests/test_results.py
```diff
@@ -26,10 +26,12 @@
 
 
 @pytest.mark.parametrize(
-    "counts_raw, expected_out",
+    "counts_raw, expected_out, remove_zeros",
     [
-        ({" 1": 474, "0": 550}, {"0": 550, "1": 474}),
-        ({"10": 479, "1 1": 13, "0 0 ": 496}, {"00": 496, "01": 0, "10": 479, "11": 13}),
+        ({" 1": 0, "0": 550}, {"0": 550}, True),
+         ({"10": 479, "1 1": 13, "0 0 ": 496}, {"00": 496, "10": 479, "11": 13}, True),
```
ryanhill1, line 32: Remove the one leading space here

qbraid/devices/tests/test_results.py
```diff
@@ -42,11 +44,12 @@
                 "110": 139,
                 "111": 594,
             },
+            False
```
ryanhill1, line 47: Add a comma after False here

ryanhill1 (changes requested):
Super close, just gotta get the formatting check passing with these two last tiny fixes

Screenshot 2023-08-15 at 1 00 03 PM

ryanhill1 (approved):

Issue #355, opened 2023-08-21
Refactor to separate core and optional dependencies


ryanhill1:
Follow-up to https://github.com/qBraid/qBraid/issues/175

As the SDK continues to add more frameworks, it is becoming too large. To make the SDK more light weight we should separate out core dependencies. E.g. if someone wants to use the SDK just to transpile circuits between two packages, they shouldn't have to install all of the other packages that they will never use.

As of [v0.4.4](https://github.com/qBraid/qBraid/releases/tag/v0.4.4) the core dependencies would be

```
cirq
numpy
requests
```

and the optional dependencies would be

```
amazon-braket-sdk
pyquil
pytket
pytket-braket
qiskit
qiskit-ibm-provider
qiskit-qasm3-import
```

These optional dependencies could then be grouped into different [`extras`](https://github.com/qBraid/qBraid/blob/a572428d2e307fdb0286098e627f4e148c15f17c/setup.cfg#L45) to allow users to optionally install them based on which framework(s) they need to convert between. 

Implementation while require a top-to-bottom refactoring. Further work is needed to plan out what exactly this new architecture would look like.

Linking further description: https://github.com/qBraid/qBraid/pull/357#pullrequestreview-1590205569

@TheGupta2012 @mmlouamri @rryoung98 @junliangtan1 @poig

EnriqueAnguianoVara:
Hi! Can I work on this?

ryanhill1:
Also related to https://github.com/qBraid/qBraid/issues/186

PR #357, opened 2023-08-22
fixes #335

EnriqueAnguianoVara:
#335

## Changes:
Separate core and optional requirements in setup.cfg and setup.py

## Details or comments:
I did not check the last general boxes as I am having some issues with both (this is my first pr here and I think the best thing is not to do things that might be wrong, minimizing the work of all ) So here I have some questions to be able to finish this PR:

-  The unit test I supposed that I need to create would be in ``qbraid/test/test_top_level``?
-  My idea to create a test would be installing and comparing if the output is 0, or find it in the installed modules. Do you have a favorite way to do it here?. 
- Somehow, when I test ``tox -e docs`` a document test, (not related at all with the things I change) send me a error that I don't passed the test. Is this an issue that is / need to be reported or somehow is my fault?

I want to thank all the people to have the patience to read and request changes (the ones I told and maybe others) in my first of many (I hope) PR in qBraid

## Checklist before merge

### General 

- [ ] ALL new features must include unit test to the test directory, codecov will check for any missing test code exist.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -30,19 +30,21 @@ packages = find:
 python_requires = >=3.9
 
 install_requires =
-    amazon-braket-sdk>=1.42.1,<1.60.0
     cirq>=1.3.0.dev20230801170401,<1.4.0
     numpy>=1.17,<1.24
+    requests>=2.28.0
+    ipython
+
+[options.extras_require]
+dependency =
+    amazon-braket-sdk>=1.42.1,<1.60.0
     pyquil>=3.5.4,<3.6.0
     pytket>=1.16,<1.19
     pytket-braket>=0.27,<0.30
     qiskit>=0.44.0
     qiskit-ibm-provider>=0.5.3
     qiskit-qasm3-import>=0.2.0,<0.3.0
-    requests>=2.28.0
-    ipython
 
-[options.extras_require]
 test =
     sympy
     pytest
--- a/setup.py
+++ b/setup.py
@@ -18,4 +18,23 @@ from setuptools import setup
 with open("qbraid/_version.py") as f:
     version = f.readlines()[-1].split()[-1].strip("\"'")
 
-setup(version=version)
+setup(
+    version=version,
+    extras_require={ 
+        # Install all optional dependencies
+        'all':['dependency'],
+        #Bra-ket
+        'amazon-braket': ['amazon-braket-sdk'],
+        #Pyquill
+        'rigetti':['pyquill'],
+        #Pytket
+        'pytket-all': ['pytket','pytket-braket'],
+        'pytket':['pytket'],
+        'pytket-braket':['pytket-braket'],
+        #Qiskit
+        'qiskit-all': ['qiskit','qiskit-ibm-provider','qiskit-qasm3-import'],
+        'qiskit':['qiskit'],
+        'qiskit-ibm-provider':['qiskit-ibm-provider'],
+        'qiskit-qasm3-import':['qiskit-qasm3-import']
+    }
+)


ryanhill1 (changes requested):
Hi @EnriqueAnguianoVara, thanks for this PR! Unfortunately, separating out the core and optional dependencies will be a little more complex than just updating these setup files.

Right now, locally, unit tests might still seem to pass with your updates because tox installs dependencies from `requirements.txt`. 

https://github.com/qBraid/qBraid/blob/a572428d2e307fdb0286098e627f4e148c15f17c/tox.ini#L14

so all of the current packages would still be installed for testing. However, the docs build instead uses a development install, which is why that was failing. To test for yourself, you can create a new clean python virtual environment, perform a development install from your development branch

```
$ pip install -e .
```
and then try to `import qbraid` from a python shell. You will find that the import fails because of multiple `ModuleNotFoundError` across the missing "optional" packages. This is because despite declaring the packages as optional in the setup files, right now, the SDK is structured in a way that importing `qbraid` directly imports each of the currently listed packages in `requirements.txt`.

To complete this issue, we need to "re-factor" (i.e. change the structure of the existing code without altering its core functionality) the SDK so that features that extend "optional" packages are only imported and available if that given package is installed. As mentioned in https://github.com/qBraid/qBraid/issues/355: 

> Implementation while require a top-to-bottom refactoring

As *one* example, in the future re-factored version, `QPROGRAM_LIBS` should only return libraries available from the core requirements, i.e.:

```python
>>> from qbraid import QPROGRAM_LIBS
>>> QPROGRAM_LIBS
['cirq',  'qasm2']
```
But if I then installed some optional dependencies with

```bash
$ pip install 'qbraid[qiskit]' 'qbraid['qasm3']
```
then the `QPROGRAM_LIBS` available via `qbraid.interface` and `qbraid.transpiler` functions would be updated accordingly:

```python
>>> from qbraid import QPROGRAM_LIBS
>>> QPROGRAM_LIBS
['cirq', 'qiskit', 'qasm2', 'qasm3']
```

Let me know if this makes any sense! If this issue is something you would still be interested in I would be happy to set up a time to discuss architectures that we meet these requirements.

EnriqueAnguianoVara:
It makes sense and I would love to do it. For setting the time of the discussion, I'm quite available tomorrow all day long ( in CEST timezone)

ryanhill1:
> It makes sense and I would love to do it. For setting the time of the discussion, I'm quite available tomorrow all day long ( in CEST timezone)

@EnriqueAnguianoVara Sounds great! Just messaged you on LinkedIn so we can setup a time

ryanhill1 (changes requested):
Thanks for this PR! Your changes to `_qprogram.py` look great. I tested the file and it seems to resolve #367.

Since we're doing the optional dependency migration in steps, the changes to `setup.cfg` will be the last thing that we do. Until the repository is completely setup for the transition, separating out these dependencies will cause build issues. So for this PR, let's just reset the changes to `setup.cfg` and `setup.py`, so we can merge the changes and close the `_qprogram.py` issue.

*Side note*: When we are ready for the transition, configuring the optional dependencies i.e. extras in `setup.cfg` is sufficient. There won't actually be any changes needed to `setup.py`.

As far as CI/CD, make sure that you read [`CONTRIBUTING.md`](https://github.com/qBraid/qBraid/blob/main/CONTRIBUTING.md) to ensure that your code passes integration tests. That file as instructions explaining how to use `tox` to test locally.

ryanhill1 (approved):

Issue #298, opened 2023-07-10
OpenQASM 3 convert to contiguous


ryanhill1:
Write function that converts an QASM 3 string circuit to use only contiguous qubit indexing i.e removes empty registers of unused qubits. Similar to [`_convert_to_contiguous_qasm()`](https://github.com/qBraid/qBraid/blob/87bd183e8f1262df7a6ea939c74c81b74d1ea050/qbraid/interface/qbraid_qasm/tools.py#L54C5-L54C32) but for qasm3 instead of qasm2, and implemented without any external dependencies, i.e. should not rely on cirq conversion as existing qasm2 implementation does.

TheGupta2012:
Can I take a look at this too @ryanhill1 ?

ryanhill1:
@TheGupta2012 absolutely!

PR #398, opened 2023-09-07
Add implementation for `contiguous_qasm_3` function

TheGupta2012:
- Issue Tag : #298 

## Changes:
New function `convert_to_contiguous_qasm3` added in `qbraid/interface/qbraid_qasm/tools.py`

## Details or comments:
- Aims at making a contiguous circuit through expansion. Added `I` gates over unused qubit indices. 
- The function is not dependent on any external libraries but uses `re` package (this may be a performance bottleneck but it works as of now) 
- The function first splits the qasm string and finds the register declarations 
- Post this, any non-gate statements are removed and qubit indices are searched for in each gate statement
- This generates a mapping of unused qubit indices and we then simply append an `I` gate for each of them
## Checklist before merge

### General 


ryanhill1 (changes requested):
Thanks for the PR, this looks great! Can you resolve merge conflicts with main so we can run workflows?

tests/interface/test_qasm3_utils.py
```diff
@@ -26,8 +26,11 @@
 from qbraid.interface.qbraid_qasm3.tools import (
     convert_to_qasm3,
     qasm3_depth,
-    qasm3_num_qubits,
-    qasm3_qubits,
+    qasm_3_num_qubits,
+    qasm_depth,
+    qasm_num_qubits,
+    qasm_qubits,
```
ryanhill1, line 32: Updating these naming conventions should fix the import errors in the tests
- `qasm_3_num_qubits` $\rightarrow$ `qasm3_num_qubits`
- `qasm_depth` $\rightarrow$ `qasm3_depth`
- `qasm_num_qubits` $\rightarrow$ `qasm3_num_qubits`
- `qasm_qubits` $\rightarrow$ `qasm3_qubits`

ryanhill1 (changes requested):
Seems like some of the function names got reverted in the merge which is what's causing the unit tests to fail.

After doing that, you can run
```
tox -e unit-tests
tox -e format-check
```
locally to make sure everything passing.

tests/interface/test_qasm3_utils.py
TheGupta2012: Thanks!

TheGupta2012:
@ryanhill1 This implementation does not account for the possibility of custom functions and looping statements which are a part of the qasm 3 grammar. 

It will be nice to have something like an `Unroller` to remove branching and then process qasm3 strings using the contiguous function. Or any other approach you had in mind works too! 

P.S. using `openqasm.parse` function seems like a pretty clean way to handle this but I'm not sure if that external dependency is something which is desired

qbraid/interface/qbraid_qasm3/tools.py
```diff
@@ -72,6 +73,127 @@ def _convert_to_contiguous_qasm3(qasmstr: QASMType, rev_qubits=False) -> QASMTyp
...
+    # 2. Identify which qubits are not used in the circuit
+    for qubit_decl in qubit_list:
+        qreg_list.append(_get_qreg(qubit_decl))
+
+    # 3. Add an identity gate for the unused qubits
+    qubit_indices = _get_unused_qubit_indices(qasmstr, qreg_list)
+    expansion_qasm = ""
+    for reg, indices in qubit_indices.items():
+        for index in indices:
+            expansion_qasm += f"i {reg}[{index}];\n"
```
ryanhill1, line 191: For the default behavior of the convert to contiguous function, we want to collapse any empty registers. Optionally, you can add the `expansion` keyword argument to the function definition, e.g.

https://github.com/qBraid/qBraid/blob/7a436d804929404bef5ec271b687ef16981f4be1/qbraid/interface/qbraid_cirq/tools.py#L139C69-L139C84

which, if set to `True`, would instead add the identity gates to the empty registers. I think that's what you're attempting to do here, but the behavior isn't quite correct. Using your example test case:

```python
from qbraid.interface.qbraid_braket.qasm import braket_from_qasm3
from qbraid.interface.qbraid_qasm3.tools import convert_to_contiguous_qasm3

qasm_test = """
OPENQASM 3.0;
gate custom q1, q2, q3{
    x q1;
    y q2;
    z q3;
}
qreg q1[2];
qubit[2] q2;
qubit[3] q3;
qubit q4;

x q1[0];
y q2[0];
z q3;
"""

circuit_in = braket_from_qasm3(qasm_test)
circuit_out = braket_from_qasm3(convert_to_contiguous_qasm3(qasm_test))
```
```python
>>> print(circuit_in)
T  : |0|
        
q0 : -X-
        
q2 : -Y-
        
q4 : -Z-
        
q5 : -Z-
        
q6 : -Z-

T  : |0|

>>> print(circuit_out)
T  : |0|
        
q0 : -X-
        
q1 : -I-
        
q2 : -Y-
        
q3 : -I-
        
q4 : -Z-
        
q5 : -Z-
        
q6 : -Z-
        
q7 : -I-

T  : |0|
```

This input circuit was 5 qubits, with two empty registers. For for contiguous expansion, it should end up as 7 qubits, but this output circuit is 8 qubits.

ryanhill1 (changes requested):
Thanks for these updates! Left a few comments throughout the code. To address your question, yes, it's totally fine and encouraged to leverage the `openqasm3` parser. We've decided to include that in the SDK's core requirements.

 As a final step, please also substitute your new function into the existing [`_convert_to_contiguous_qasm3()`](https://github.com/qBraid/qBraid/blob/7a436d804929404bef5ec271b687ef16981f4be1/qbraid/interface/qbraid_qasm3/tools.py#L68-L72) function, to replace the current qiskit implementation, and make sure all the test cases impacted by that still pass.

Thanks, looks great!!! Let me know if I can help or answer any other questions.

qbraid/interface/qbraid_qasm3/tools.py
```diff
@@ -72,6 +73,127 @@ def _convert_to_contiguous_qasm3(qasmstr: QASMType, rev_qubits=False) -> QASMTyp
...
+
+def convert_to_contiguous_qasm3(qasmstr: str) -> QASMType:
+    """Converts OpenQASM 3 string to contiguous qasm3 string with gate expansion"""
+
+    # SCOPE : no loops, no functions at the moment
+
+    # pylint: disable=import-outside-toplevel
+
+    # 1. Analyse the qasm3 string for qubits and bit declarations
+    qubit_list = _qasm_qubit_decl(qasmstr)
```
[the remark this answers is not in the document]
TheGupta2012: That's very helpful, thanks!

qbraid/interface/qbraid_qasm3/tools.py
TheGupta2012: Aren't there a total of 8 qubits? 
q1 -> 2 qubits
q2 -> 2 qubits
q3 -> 3 qubits 
q4 -> 1 qubit

tests/interface/test_qasm3_utils.py
```diff
@@ -133,6 +134,30 @@ def test_qasm3_random_with_known_seed():
     _check_output(circuit, out__expected)
 
 
+def test_convert_to_contiguous_qasm_3():
+    """Test conversion of qasm3 to contiguous qasm3"""
+    qasm_test = """
```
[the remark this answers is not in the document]
TheGupta2012: Yep, that's the case. If you include the header it parses correctly, I'll update the test to include it

qbraid/interface/qbraid_qasm3/tools.py
TheGupta2012: One question arises here @ryanhill1 , if a register is partially used do we redeclare it as a register with a size equal to the used qubits or just let it be? 
Example - 
```
qubit[3] q1;
qubit[2] q2;
cx q1[0], q2[0];
cx q1[1], q2[1];
```
should be changed to - 
```
qubit[2] q1; // changed size
qubit[2] q2;
cx q1[0], q2[0];
cx q1[1], q2[1];
```

TheGupta2012: The approach I see is that if `expansion` is not set, we remove the unused quantum registers i.e. where no qubits were accessed.

Then, we return the expansion of the leftover registers (with the same size declarations of partially used registers, just `I` gates added to unused ones)

ryanhill1 (approved):

qbraid/interface/qbraid_qasm3/tools.py
ryanhill1: Yeah that's correct. And yes, we would re-declare the register with the adjusted size in that case.

Issue #409, opened 2023-09-13
OpenQASM 3 reverse qubit ordering function


ryanhill1:
Implement function `reverse_qubit_order` in

https://github.com/qBraid/qBraid/blob/79306c1be8e9d9dff7c4cdc4c3247467f58993a7/qbraid/programs/qasm3.py#L246

*Optional Bonus:* Extend function not just reverse qubit ordering, but transform circuit to new qubit indexing specified by an input qubit mapping dictionary.

TheGupta2012:
@ryanhill1 can I work on this and also fix second half of #403 ?

ryanhill1:
@TheGupta2012 Yup absolutely!

PR #433, opened 2023-10-04
Add `reverse_qubit_order` method for qasm3

TheGupta2012:
Fixes Issue #409 

## Changes
Added new methods `reverse_qubit_order` and `apply_qubit_mapping` to the `OpenQasm3Program` class. 
The `qbraid/interface/qbraid_qasm3/tools.py` is also updated to include the `reverse_qubit_order` function for reversing qubit without going through the circuit wrapper on the user side. 

## Details
Qubit mapping structure has been decided as follows - 

```python3 
# for remapping 2 registers
mapping = {
    'q' : { 0 : 1, 1 : 0}, 
    'q2' : { 0 : 2, 1: 0, 2 : 1}
}
```
More information is specified in the code. 

Added unit tests to cover different qubit mapping cases and catch incorrect mapping exceptions. 

### General 


TheGupta2012:
@ryanhill1 this unrelated test was failing in my local too, any advice?

ryanhill1:
@TheGupta2012 Yeah the cyclic import issue should be fixed with latest commits to main. But your dev branch is currently about ~10 commits behind main. So if you pull the latest main branch and merge into your dev branch that should fix it. If it doesn't I can take a second look.

TheGupta2012:
Don't think I'm able to catch this, I had pulled from upstream and linters test passes in local - 
```bash
(qbraid-env) harshit@harshit-Aspire-A315-55G:~/Desktop/college/quantum/qBraid $ tox -e linters 
linters: commands[0]> isort . qbraid tools tests
Skipped 8 files
linters: commands[1]> black qbraid tools tests --exclude '/(tests/transpiler/cirq_qasm|qbraid/transpiler/cirq_qasm)/'
All done! ✨ 🍰 ✨
145 files left unchanged.
linters: commands[2]> python tools/verify_headers.py qbraid tests tools --fix

All done! ✨ 🚀 ✨
0 headers fixed, 150 headers left unchanged.
  linters: OK (5.47=setup[0.15]+cmd[2.50,2.78,0.04] seconds)
  congratulations :) (5.51 seconds)
```

ryanhill1 (changes requested):
Ok, I now see where the issue is. It's because of the

```python
from qbraid.programs.qasm3 import OpenQasm3Program
``` 

line that you added to `qbraid.interface.qbraid_qasm3.tools`. Right now `qbraid.programs` already imports the circuit wrapper functions from `qbraid.interface`, so the interface module can't simultaneously import the programs module, or else you get the circular import error.

But this is besides the point, because we don't actually need the qasm3 `reverse_qubit_order` function in the interface module, because you've already implemented it in the qasm3 circuit wrapper. If you noticed, I removed all of the `reverse_qubit_order` and `convert_to_contiguous` function from the interface module and moved them into their respective circuit wrapper classes in the programs module. This makes for a much cleaner and more pythonic package structure

So to fix the issue, remove the `OpenQasm3Program` import and the `reverse_qubit_order` function from the interface qasm3 tools file. And then change your test cases to test against the `OpenQasm3Program.reverse_qubit_order` instead of the version imported from interface.

ryanhill1 (approved):

PR #446, opened 2023-10-17
Add qiskit <-> braket converter via `openqasm3`

### LINKED ISSUE #388: Qiskit <--> Braket direct conversion via OpenQASM 3
Currently, transpiling circuits between Qiskit and Amazon Braket requires going through Cirq as an intermediate conversion. Now that Amazon Braket supports converting circuits to/from OpenQASM 3, we can setup a direct link for conversions between qiskit and braket. Will eventually integrate with https://github.com/qBraid/qBraid/issues/359

**Qiskit $\leftrightarrow$ Amazon Braket**

*Current*:

qiskit $\leftrightarrow$ qasm2 $\leftrightarrow$ cirq $\leftrightarrow$ braket

*Desired*:

qiskit $\leftrightarrow$ qasm3 $\leftrightarrow$ braket

TheGupta2012:
Fixes #388 

## Changes:
Adds a new folder for the `qiskit` to `braket` conversions via `openqasm3`

Currently, we use a fallback mechanism for the conversion. If the conversion via `openqasm3` fails, we fall back to the `cirq` path.
### General 


ryanhill1 (changes requested):
This looks like some great progress! A couple things:

1. We want to add the new conversion to the`QuantumProgram.transpile` method:

https://github.com/qBraid/qBraid/blob/888cc44541a9d0daabb418c027e81ed370d0b4f2/qbraid/programs/abc_program.py#L178

So if someone calls the transpile method, if `self.package` and `conversion_type` are both either qiskit or braket, then we can do a lazy import (i.e. inside the function itself b/c qiskit and braket aren't core dependencies) of the qiskit/braket conversion function, and use that first instead of the convert to / from cirq.

This will actually clean up the code a bit, because it will allow us to remove the `convert_from_cirq(convert_to_cirq(...` and `convert_from_cirq(convert_to_cirq(...` fallbacks in the try except block in your qiskit_to_braket conversions file. We can just allow the exception to be raised there. And instead, catch the exception in transpile, and then just allow the operation to proceed as it normally would.

2. It seems from the unit tests error that Amazon Braket creates OpenQASM 3 strings with the assumption that `iswap` is pre-defined gate, whereas qiskit does not. So for a case like this, we'll have to implement something similar to 

https://github.com/qBraid/qBraid/blob/888cc44541a9d0daabb418c027e81ed370d0b4f2/qbraid/interface/qbraid_qasm/qasm_preprocess.py#L116

but for qasm3. It doesn't have to be as robust to start out. It can just be a function that checks for any `iswap` gates in the input braket OpenQASM 3 string, and if it contains one, then you just insert the iswap gate def into the top of the openqasm string, e.g.

```python
OPENQASM 3;
include "stdgates.inc";
bit[2] c;
qubit[2] q;
iswap q[0], q[1];
```
would turn into 

```python
OPENQASM 3;
include "stdgates.inc";
gate iswap _gate_q_0, _gate_q_1 {
  s _gate_q_0;
  s _gate_q_1;
  h _gate_q_0;
  cx _gate_q_0, _gate_q_1;
  cx _gate_q_1, _gate_q_0;
  h _gate_q_1;
}
bit[2] c;
qubit[2] q;
iswap q[0], q[1];
```

after being passed through. Then the iswap gates should pass in the conversion. Other edge cases may come up as well, so it would be great to explicitly write some tests for those as they appear. Happy to provide any help I can!

TheGupta2012:
Hi @ryanhill1, still a WIP but have started to incorporate the changes we discussed in the call. Let me know what you think of them.

ryanhill1 (approved):

PR #694, opened 2024-06-27
more codecov!

rjain37:
97.5%


qbraid/transpiler/conversions/cirq/cirq_quil_output.py
ryanhill1, line 1: Not a huge fan of substituting inconsistent return statements for +1 line in code coverage. I would rather leave that last `return None` in for code quality sake. If that line is "unreachable", then the function can probably be refactored in a way that preserves consistent return statements.

ryanhill1 (changes requested):
Looks great! Just the one small requested change in quil output return statements, see comment

ryanhill1 (approved):

PR #697, opened 2024-07-01
codecoverage -> 98.5%

rjain37:


qbraid/runtime/oqc/device.py
ryanhill1, line 1: If the device isn't found in `self._client.get_qpus()` instead telling the user that it's `OFFLINE` we probably want to continue to return status `UNAVAILABLE` or raise a `ResourceNotFoundException` or other error. Because in this case we don't actually know the device status because it wasn't in the list of available devices given by the client.

ryanhill1 (changes requested):
LGTM just that one small edit to the oqc device status

ryanhill1 (approved):

PR #707, opened 2024-07-08
update TargetProfile

rjain37:
changed targetprofile from collections.Mapping to pydantic.BaseModel

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,8 @@ dependencies = [
     "rustworkx>=0.14.2",
     "numpy>=1.17,<2.1",
     "openqasm3[parser]>=0.4.0,<1.1.0",
-    "qbraid-core>=0.1.13"
+    "qbraid-core>=0.1.13",
+    "pydantic>2.0.0"
 ]
 requires-python = ">=3.9,<3.13"
 
--- a/qbraid/programs/spec.py
+++ b/qbraid/programs/spec.py
@@ -48,12 +48,12 @@ class ProgramSpec:
         return self._native
 
     def __str__(self) -> str:
-        return f"ProgramSpec for {self.alias} {self._program_type.__name__} type."
+        return f"ProgramSpec({self._program_type.__name__}, {self.alias})"
 
     def __repr__(self) -> str:
         return (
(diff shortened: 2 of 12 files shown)


ryanhill1 (changes requested):
Looks good so far! The motivation behind the conversion to `pydantic` is with the hope that we would be able to take advantage of their built-in [validators](https://docs.pydantic.dev/latest/concepts/validators/), and save ourselves the hassle of all the extensive `def _validate_<field>` type checks that we are currently doing manually. For example:

```python
from pydantic import BaseModel, validator
from typing import Optional, List, Set
from qbraid.programs import ProgramSpec
from .enums import DeviceActionType, DeviceType

class TargetProfile(BaseModel):
    device_id: str
    device_type: DeviceType
    action_type: Optional[DeviceActionType] = None
    basis_gates: Optional[Set[str]] = None
    num_qubits: Optional[int] = None
    program_spec: Optional[ProgramSpec] = None
    provider_name: Optional[str] = None

    @validator('device_type', 'action_type', pre=True)
    def ensure_enum(cls, v, field):
        if isinstance(v, str):
            return field.type_(v.upper())
        return v

    @validator('basis_gates', pre=True)
    def convert_basis_gates_to_set(cls, v):
        if v is not None:
            return set(gate.lower() for gate in v)
        return v

    # etc...
```

With `pydantic`, we can structure the `TargetProfile` more like a data class, get rid of the `__init__` method, and use the `@validator` decorator to streamline type validation and error handling with more concise syntax. There are number of other built-in options as well that may be better suited e.g. [annotated validators](https://docs.pydantic.dev/latest/concepts/validators/#annotated-validators), [validators for default values](https://docs.pydantic.dev/latest/concepts/validators/#validation-of-default-values), [field validators](https://docs.pydantic.dev/latest/concepts/validators/#field-validators), and much more. Take a look, and try to figure out how we can optimize our approach using `pydantics` built-in functionality in a way that requires the least amount to overhead to add new fields going forward.

ryanhill1 (changes requested):
Leaving a request comment

ryanhill1 (approved):

PR #710, opened 2024-07-11
Azure runtime V3

pranetksharma:
# Changes
Completed major fixes to Azure Runtime, including full code coverage for the tests and bug fixes for the Devices and Provider class. The dependencies might still need to be checked.

## PR Checklist

- [x] I have read `CONTRIBUTING.md`.
- [x] All new code includes corresponding unit tests and satisfies code coverage.
- [x] API Reference docstrings have been updated for new features and changes in functionality.
- [x] All integration tests, including remote tests, are passing.
- [x] (Optional) My name and affiliation are added to `CITATION.cff`.

### DIFF
--- a/qbraid/runtime/azure/__init__.py
+++ b/qbraid/runtime/azure/__init__.py
@@ -1,0 +1,30 @@
+# Copyright (C) 2024 qBraid
+#
+# This file is part of the qBraid-SDK
+#
+# The qBraid-SDK is free software released under the GNU General Public License v3
+# or later. You can redistribute and/or modify it under the terms of the GPL v3.
+# See the LICENSE file in the project root or <https://www.gnu.org/licenses/gpl-3.0.html>.
+#
+# THERE IS NO WARRANTY for the qBraid-SDK, as per Section 15 of the GPL v3.
+
+"""
+Mdule submiting and managing jobs through the Azure Quantum API.
+
+.. currentmodule:: qbraid.runtime.azure
+
+Classes
+--------
+
+.. autosummary::
+   :toctree: ../stubs/
+
+	AzureSession
+	AzureQuantumProvider
+    AzureQuantumDevice
+
+"""
+from .device import AzureQuantumDevice
+from .provider import AzureQuantumProvider, AzureSession
+
+__all__ = ["AzureSession", "AzureQuantumProvider", "AzureQuantumDevice"]
--- a/qbraid/runtime/azure/device.py
+++ b/qbraid/runtime/azure/device.py
@@ -1,0 +1,52 @@
+# Copyright (C) 2024 qBraid
+#
+# This file is part of the qBraid-SDK
+#
+# The qBraid-SDK is free software released under the GNU General Public License v3
+# or later. You can redistribute and/or modify it under the terms of the GPL v3.
+# See the LICENSE file in the project root or <https://www.gnu.org/licenses/gpl-3.0.html>.
+#
+# THERE IS NO WARRANTY for the qBraid-SDK, as per Section 15 of the GPL v3.
+
+"""
+Module defining Azure Quantum device class
+
+"""
+from typing import TYPE_CHECKING
+
+from qbraid.runtime.device import QuantumDevice
+
+if TYPE_CHECKING:
+    import qbraid.runtime
+    from qbraid.runtime.azure.provider import AzureSession
+
+
+class AzureQuantumDevice(QuantumDevice):
+    """Azure quantum device interface."""
+
+    def __init__(
+        self,
+        profile: "qbraid.runtime.TargetProfile",
+        session: "qbraid.runtime.azure.provider.AzureSession",
+    ):
+        super().__init__(profile=profile)
+        self._session = session
+
+    @property
+    def session(self) -> "qbraid.runtime.azure.provider.AzureSession":
+        """Return the Azure session."""
+        return self._session
+
+    def status(self) -> "qbraid.runtime.DeviceStatus":
+        """Return the current status of the Azure device."""
+        return self.profile.get("status")
+
+    def submit(
+        self, run_input, name, provider, backend, qubits, **kwargs
+    ):  # pylint: disable=too-many-arguments, disable=arguments-differ
+        """Submit a job to the Azure device."""
+        # is_single_input = not isinstance(run_input, list)
+        # run_input = [run_input] if is_single_input else run_input
+
+        job = self._session.create_job(run_input, name, provider, backend, qubits)
+        return job
--- a/qbraid/runtime/azure/provider.py
+++ b/qbraid/runtime/azure/provider.py
@@ -1,0 +1,494 @@
+# Copyright (C) 2024 qBraid
+#
+# This file is part of the qBraid-SDK
+#
+# The qBraid-SDK is free software released under the GNU General Public License v3
+# or later. You can redistribute and/or modify it under the terms of the GPL v3.
+# See the LICENSE file in the project root or <https://www.gnu.org/licenses/gpl-3.0.html>.
+#
+# THERE IS NO WARRANTY for the qBraid-SDK, as per Section 15 of the GPL v3.
+
+"""
+Module defining Azure session and provider classes
+
+"""
+from typing import Any
+from uuid import uuid4
+
+import openqasm3
+from azure.storage.blob import BlobServiceClient, ContentSettings
+from qbraid_core._import import LazyLoader
+from qbraid_core.sessions import Session
+
+from qbraid.programs.spec import ProgramSpec
+from qbraid.runtime.enums import DeviceType
+from qbraid.runtime.profile import TargetProfile
+from qbraid.runtime.provider import QuantumProvider
+
+from .device import AzureQuantumDevice
+
+pyquil = LazyLoader("pyquil", globals(), "pyquil")
+
+
+class AzureSession(Session):
+    """Azure session class."""
+
+    def __init__(
+        self,
+        client_id: str,
+        client_secret: str,
+        tenant_id: str,
+        location_name: str,
+        subscription_id: str,
+        resource_group_name: str,
+        workspace_name: str,
+        storage_account: str,
+        connection_string: str,
+    ):  # pylint: disable= too-many-arguments
+
+        self.client_id = client_id
+        self.client_secret = client_secret
+        self.tenant_id = tenant_id
+        self.location_name = location_name
+        self.subscription_id = subscription_id
+        self.resource_group_name = resource_group_name
+        self.workspace_name = workspace_name
+        self.storage_account = storage_account
+        self.connection_string = connection_string
+        self.backends = {
+            "Syntax Checker": "quantinuum.sim.h1-1sc",
+            "Emulator": "quantinuum.sim.h1-1e",
+            "QPU": "quantinuum.qpu.h1-1",
+        }  # will need to add for pyquil
+        self.formats = {"quantinuum": ["honeywell.openqasm.v1", "honeywell.quantum-results.v1"]}
+
+        self.jobs = {}
+        self.session = Session()
+        self._helpers = AzureHelperFunctions()
+
+        self.base_url = (f"https://{location_name}.quantum.azure.com",)
+
+        super().__init__(base_url=self.base_url)
+
+    def get_devices(self) -> list[str, dict[str, Any]]:
+        """Get all Azure Quantum devices."""
+        quantum_access_token = self._helpers.quantum_access_token(
+            self.client_id, self.client_secret, self.tenant_id, self.session
+        )
+
+        url = (
+            f"https://{self.location_name}.quantum.azure.com/subscriptions/{self.subscription_id}"
+            f"/resourceGroups/{self.resource_group_name}/providers/Microsoft.Quantum/"
+            f"workspaces/{self.workspace_name}/providerStatus?api-version=2022-09-12-preview"
+        )
+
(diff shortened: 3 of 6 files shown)


requirements-dev.txt
```diff
@@ -5,6 +5,7 @@ pennylane<0.37
 pyquil>=4.4,<4.11; python_version < "3.12"
 pytket>=1.26,<1.30
 qiskit>=0.44,<1.2
+azure-quantum
```
ryanhill1, line 8: Move under `# optional runtime dependencies` section and add both azure-quantum and azure-storage-blob deps with min / max version bounds, same as you gave it `pyproject.toml`.

qbraid/runtime/azure/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+
+Classes
+--------
+
+.. autosummary::
+   :toctree: ../stubs/
+
+	AzureSession
+	AzureQuantumProvider
+    AzureQuantumDevice
```
ryanhill1, line 24: Indent 3 spaces to align with `:toctree:`, kind of weird. sphinx, idk

qbraid/runtime/azure/device.py
```diff
@@ -0,0 +1,52 @@
...
+    def status(self) -> "qbraid.runtime.DeviceStatus":
+        """Return the current status of the Azure device."""
+        return self.profile.get("status")
+
+    def submit(
+        self, run_input, name, provider, backend, qubits, **kwargs
+    ):  # pylint: disable=too-many-arguments, disable=arguments-differ
+        """Submit a job to the Azure device."""
+        # is_single_input = not isinstance(run_input, list)
+        # run_input = [run_input] if is_single_input else run_input
```
ryanhill1, line 49: From this, guessing this implementation only supports single jobs, not batch jobs? If so, that's totally fine. But since the type hint given by the parent class says `run_input` can be a list, will want to raise an exception here if list inputs are not supported to clearly communicate that to the user

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        self.client_secret = client_secret
+        self.tenant_id = tenant_id
+        self.location_name = location_name
+        self.subscription_id = subscription_id
+        self.resource_group_name = resource_group_name
+        self.workspace_name = workspace_name
+        self.storage_account = storage_account
+        self.connection_string = connection_string
+        self.backends = {
+            "Syntax Checker": "quantinuum.sim.h1-1sc",
```
ryanhill1, line 59: *Opinion*: avoid spaces in dictionary keys when possible

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+
+        self.jobs = {}
+        self.session = Session()
+        self._helpers = AzureHelperFunctions()
+
+        self.base_url = (f"https://{location_name}.quantum.azure.com",)
+
+        super().__init__(base_url=self.base_url)
+
+    def get_devices(self) -> list[str, dict[str, Any]]:
```
ryanhill1, line 73: Return type of this function looks like a `list[dict[str, Any]]`, not a `list[str, dict[str, Any]]`

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+    def get_device(self, device_id: str) -> dict[str, Any]:
+        """Get a specific Azure Quantum device."""
+        devices = self.get_devices()
+        device = None
+
+        for i in devices:
+            if i.get("id") == device_id:
+                device = i
+
+        return device
```
ryanhill1, line 103: If ID matches, return right away. If you get through the entire list and no IDs matched, raise a `qbraid.runtime.ResourceNotFoundError`, or similar.

Returning `None` when resource is not found will just figuratively "kick the rock down the road" as to when the exception is raised.

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+            containerSasUri,
+            inputSasUri,
+            outputSasUri,
+            job_id,
+            job_name=name,
+            input_data_format=self.formats[provider][0],
+            provider=provider,
+            backend=self.backends[backend],
+            num_qubits=qubits,
+            num_shots=100,
```
ryanhill1, line 157: let's let the user specify the number of shots as a kwarg

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+            job_payload,
+            job_headers,
+            self.location_name,
+            self.subscription_id,
+            self.resource_group_name,
+            self.workspace_name,
+            self.session,
+        )
+
+        self.jobs[job_id] = [job_headers, container_name]
```
ryanhill1, line 173: Using class attributes as global storage is not advisable because it can lead to unexpected behavior due to shared state among instances. Instead we want to use encapsulation to make  class's behavior more predictable i.e. Whatever data is needed for the next step should just be passed in the return.

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        header = self.jobs[job_id][0]
+        url = (
+            f"https://{self.location_name}.quantum.azure.com/subscriptions/"
+            f"{self.subscription_id}/resourceGroups/{self.resource_group_name}/"
+            f"providers/Microsoft.Quantum/workspaces/{self.workspace_name}/jobs/"
+            f"{job}?api-version=2022-09-12-preview"
+        )
+
+        response = self.session.delete(url, headers=header)
+        print("Job has been cancelled.")
```
ryanhill1, line 209: Use `logger.info` instead of print. Other developers who use the package can configure log-level to show which stdout / stderr messages they want to see. print bypasses that and prints to the console automatically without any easy way for people to shut it off

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+            client_id=client_id,
+            client_secret=client_secret,
+            tenant_id=tenant_id,
+            location_name=location_name,
+            subscription_id=subscription_id,
+            resource_group_name=resource_group_name,
+            workspace_name=workspace_name,
+            storage_account=storage_account,
+            connection_string=connection_string,
+        )
```
ryanhill1, line 239: Use `**kwargs` and pass directly through or require `AzureSession` as argument to avoid redundancy

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+
+        queue_time = data.get("averageQueueTime")
+        status = data.get("status")
+
+        return TargetProfile(
+            device_id=data.get("id"),
+            device_type=device_type,
+            program_spec=program_spec,
+            queue_time=queue_time,
+            status=status,
```
ryanhill1, line 264: `TargetProfile` is meant for static values that describe the characteristics of a quantum device and the runtime configurations that dictate how and if a program can be submitted to that device.

Values such as `queue_time` and `status` are dynamic in nature. They might be different today then they are tomorrow. Therefore, we should retrieve these values via API request in real-time from inside the device class.

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+
+        headers = {"Content-Type": "application/x-www-form-urlencoded"}
+        url = f"https://login.microsoftonline.com/{tenant_id}/oauth2/v2.0/token"
+        r = session.post(url, headers=headers, data=data, verify=True)
+        quantum_access_token = r.json()["access_token"]
+
+        return quantum_access_token
+
+    def storage_access_token(
+        self, client_id: str, client_secret: str, tenant_id: str, session: Any
```
ryanhill1, line 305: is `session = Any` an `AzureSession`? If so, let's note that in the type hint

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+                if machine["id"][:4] != "ionq":
+                    devices_dict[machine["id"]] = {
+                        "status": machine["currentAvailability"],
+                        "isAvailable": machine["currentAvailability"] == "Available",
+                        "nextAvailable": None,
+                        "availablilityCD": "",
+                        "averageQueueTime": machine["averageQueueTime"],
+                    }
+
+        return devices_dict
```
ryanhill1, line 338: status and availability data should be dynamically retrieved from methods within the device class, not statically loaded. Also, you do not need to conform the format of the dictionary to our API schema. Our schemas can and will change. Uses the data structures that make the most sense in the context.

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        session: Any,
+    ) -> list[str]:
+        """Create a new container for the job."""
+        storage_access_token = token
+
+        # NOTE: This step can be replaced with any code that qBraid has to generate job IDs,
+        # but even the official Azure documentation uses UUIDs for job IDs
+
+        job_id = str(uuid4())
+        container_name = f"job-{job_id}"  # name of the container to create
```
ryanhill1, line 355: Are you sure they allow you to assign your own job ID? what if it's not unique? I'm skeptical about this...

The container_name makes more sense to me, that's just where you plan to store the job data, kind of like an s3 bucket. So makes sense they would let you name that on your own.

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+            f"/blobServices/default/containers/{container_name}?api-version=2022-09-01"
+        )
+
+        session.put(url, json={}, headers=container_headers)
+        return [job_id, container_name]
+
+    # pylint: disable=too-many-arguments
+    @staticmethod
+    def create_job_routes(
+        session: str,
```
ryanhill1, line 374: session of type string seems wrong

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        container: str,
+        qasm: str,
+        connection_string: str,
```
ryanhill1, line 377: if you have a type hint you don't need to include the type in the variable name

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        location_name: str,
+        subscription_id: str,
+        resource_group_name: str,
+        workspace_name: str,
+        quantum_access_token: str,
+    ) -> list[str]:
+        """Create the routes for the job."""
+        containerName = container
```
ryanhill1, line 385: why switch to camel case?

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        qasm_str = qasm
```
ryanhill1, line 386: why rename qasm variable?

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+            "itemType": "Job",
+            "name": job_name,
+            "providerId": provider,
+            "target": backend,  # changeable
+            "metadata": {
+                "qiskit": "True",
+                "name": job_name,
+                "num_qubits": num_qubits,
+                "metadata": "null",
+                "meas_map": "[0]",
```
ryanhill1, line 454: use `json.dumps` to convert python types to string format. Don't need to encode as strings yourself

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+        output_uri: str,
+        job_id: str,
+        job_name: str,
+        input_data_format: str,
+        provider: str,
+        backend: str,
+        num_qubits: str,
+        num_shots: int,
+        total_count: int,
+        output_data_format: str,
```
ryanhill1, line 438: Are you sure these are all required request parameters? Or will Azure populate some of these with default values if not provided? If so, we can convert whichever are not explicitly required into optional arguments

qbraid/runtime/azure/provider.py
```diff
@@ -0,0 +1,494 @@
...
+            f"https://{location_name}.quantum.azure.com/subscriptions/{subscription_id}"
+            f"/resourceGroups/{resource_group_name}/providers/Microsoft.Quantum/workspaces/"
+            f"{workspace_name}/jobs/{job_id}?api-version=2022-09-12-preview"
+        )
+
+        response = session.put(url, json=payload, headers=job_header)
+        return response.json()
+
+    @staticmethod
+    def get_output_data(container_name: str, connection_string: str) -> dict[str, Any]:
```
ryanhill1, line 487: This function probably belongs in a job class. See: https://docs.qbraid.com/sdk/user-guide/runtime/components#building-pipelines

> [QuantumProvider](https://docs.qbraid.com/sdk/user-guide/runtime/new-provider#quantumprovider): Converts device metadata into accessible Python objects, facilitating interaction with diverse quantum computing resources.
[QuantumDevice](https://docs.qbraid.com/sdk/user-guide/runtime/new-provider#quantumdevice): Describes the unique parameters and operational settings necessary for executing quantum programs on specific hardware.
[QuantumJob](https://docs.qbraid.com/sdk/user-guide/runtime/new-provider#quantumjob): Represents the transitional states of quantum programs, managing both ongoing and completed quantum computations.
[QuantumJobResult](https://docs.qbraid.com/sdk/user-guide/runtime/new-provider#quantumresult): Captures and structures the output from quantum computations, including histograms and raw data, for analysis and interpretation.

That's the next phase of this project.

ryanhill1 (changes requested):
Thanks for these updates! Progress looks great so far. Left some comments. Happy to discuss further.

PR #711, opened 2024-07-11
conversion graph weighting fix

rjain37:
made a small mistake; 

current deployment uses weights $w = \dfrac{1}{a}$ where $a \in [0, 1.0]$ is the accuracy of the conversion. this is incorrect:

consider two paths from $A \to C$: $A \to B_1 \to C$ and $A\to B_2 \to C$. let the first path have accuracies of $\dfrac 14$ and $\dfrac 16$ whereas the two conversions in the second half both have accuracy $\dfrac 15$.  that is, the weights are 4, 6 and 5, 5 respectively.  the current implementation will consider these paths the same weight, as they both sum to 10. however, $\dfrac{1}{24} > \dfrac{1}{25}$, so we would want to prefer the first path over the second.

the fix just takes the logarithms of the current weight. this works because:
1. djikstra's is an algorithm that finds the minimum ***sum*** of edge weights
2. $\log$ is an increasing function

### DIFF
--- a/qbraid/transpiler/edge.py
+++ b/qbraid/transpiler/edge.py
@@ -16,6 +16,8 @@ import importlib
 import inspect
 from typing import TYPE_CHECKING, Any, Callable, Optional, Union
 
+import numpy as np
+
 from qbraid.programs import QPROGRAM_REGISTRY, get_program_type_alias
 
 if TYPE_CHECKING:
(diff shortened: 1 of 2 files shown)


rjain37:
image

ryanhill1 (approved):

PR #723, opened 2024-07-26
Add Azure Quantum Provider to qBraid Runtime

pranetksharma:
## Summary of changes

- Azure quantum provider + device + job + result pipeline added to `qbraid.runtime`
- ref: https://github.com/microsoft/azure-quantum-python


ryanhill1 (changes requested):
1. I pushed some edits that eliminate the need for the `AzureClient` class. Please update the tests to account for this.
2. Make sure that `AzureResult.measurements()` correctly returns a `np.ndarray`, and that `AzureResult.measurement_counts()` returns a dictionary of histogram data that can be plotted with `qbraid.visualization.plot_histogram()`and `qbraid.visualization.plot_distribution()`. See [docs](https://docs.qbraid.com/sdk/user-guide/visualization#plot-experimental-results).
3. `azure.quantum.Job.get_results()` is not guaranteed to return a dictionary. If there are errors decoding the data, it will return the raw payload. So let's make sure that we handle that case accordingly, e.g. by raising an exception.

ryanhill1 (commented):
@pranetksharma Did you have a chance to try to submit any actual jobs? When I tried to submit an OpenQASM program to `quantinuum.sim.h1-1sc` I got 

```
ValueError: Connection string is either blank or malformed.
```

Did that happen for you?

pranetksharma:
Hmm, this may be a credential issue. Let me look into it and get back to
you!

ryanhill1 (dismissed):

ryanhill1 (approved):

PR #755, opened 2024-09-17
Add device cache for `QuantumProvider`

### LINKED ISSUE #753: Add a cache for `get_devices` methods for runtime providers
### Feature Description

Currently, the runtime providers' `get_devices` method calls the API every single time to see what devices are available. This can be cached with a **time to live** attribute for each provider (in the order of minutes). The motivation comes from the fact that device availability might not vary a lot over a short period of time and also that the amazon provider takes a lot of time! 

### Implementation (Optional)

The implementation for amazon is present in the [`runtime.result` modification PR](https://github.com/qBraid/qBraid/pull/747/commits/c335bde1d8bff5a29e1a98d403b310bf59485351#file-qbraid-runtime-braket-provider-py-L164). We need to adapt this to each provider.

TheGupta2012:
## Summary of changes

Fixes #753 by adding a device cache to quantum provider. Note, the time to live for cache can be modified for each of the provider; for now its set to 120 seconds.


ryanhill1 (changes requested):
I like this! However, I think it can be done a little more cleanly using decorators. For example:

```python
import functools
import hashlib
import json
import time
from abc import ABC, abstractmethod


def _generate_cache_key(func_name: str, args: tuple, kwargs: dict) -> str:
    """Generate a cache key based on function name, args, and kwargs."""
    key_data = {"func_name": func_name, "args": args, "kwargs": kwargs}
    key_str = json.dumps(key_data, sort_keys=True)  # Ensure consistent key order
    return hashlib.sha256(key_str.encode()).hexdigest()


def cache_results(ttl: int = 120):
    """
    A decorator to cache the results of get_device or get_devices methods.

    Args:
        ttl (int): Time-to-live for the cache in seconds.
    """

    def decorator(func):
        cache = {}

        @functools.wraps(func)
        def wrapper(self, *args, **kwargs):
            # Generate a unique cache key based on method name, args, and kwargs
            key = _generate_cache_key(func.__name__, args, kwargs)
            current_time = time.time()

            # Check if the result is in cache and still valid
            if key in cache:
                cached_result, timestamp = cache[key]
                if (current_time - timestamp) < ttl:
                    return cached_result

            # Call the actual method and cache the result
            result = func(self, *args, **kwargs)
            cache[key] = (result, current_time)
            return result

        return wrapper

    return decorator


class QuantumProvider(ABC):
    """
    This class is responsible for managing the interactions and
    authentications with various Quantum services.

    """

    def save_config(self, **kwargs):
        """Saves account data and/or credentials to the disk."""
        raise NotImplementedError

    @abstractmethod
    @cache_results()
    def get_devices(self, **kwargs):
        """Return a list of backends matching the specified filtering."""

    @abstractmethod
    @cache_results()
    def get_device(self, device_id: str):
        """Return quantum device corresponding to the specified device ID."""

    def __eq__(self, other):
        """Equality comparison.

        By default, it is assumed that two `QuantumProviders` from the same class are
        equal. Subclassed providers can override this behavior.
        """
        return type(self).__name__ == type(other).__name__

```

And then, the `BraketProvider`, for example, would look something like this:

```python
class BraketProvider(QuantumProvider):
    """
    This class is responsible for managing the interactions and
    authentications with the AWS services.

    Attributes:
        aws_access_key_id (str): AWS access key ID for authenticating with AWS services.
        aws_secret_access_key (str): AWS secret access key for authenticating with AWS services.
    """

    REGIONS = ("us-east-1", "us-west-1", "us-west-2", "eu-west-2")

    def __init__(
        self, aws_access_key_id: Optional[str] = None, aws_secret_access_key: Optional[str] = None
    ):
        """
        Initializes the QbraidProvider object with optional AWS credentials.

        Args:
            aws_access_key_id (str, optional): AWS access key ID. Defaults to None.
            aws_secret_access_key (str, optional): AWS secret access token. Defaults to None.
        """
        super().__init__()
        self.aws_access_key_id = aws_access_key_id or os.getenv("AWS_ACCESS_KEY_ID")
        self.aws_secret_access_key = aws_secret_access_key or os.getenv("AWS_SECRET_ACCESS_KEY")

    ....

    @cache_results(ttl=120)
    def get_devices(
        self,
        aws_session: Optional[braket.aws.AwsSession] = None,
        statuses: Optional[list[str]] = None,
        **kwargs,
    ) -> list[qbraid.runtime.braket.BraketDevice]:
        """Return a list of backends matching the specified filtering."""
        aws_session = self._get_aws_session() if aws_session is None else aws_session
        statuses = ["ONLINE", "OFFLINE"] if statuses is None else statuses
        aws_devices = AwsDevice.get_devices(aws_session=aws_session, statuses=statuses, **kwargs)
        return [
            BraketDevice(profile=self._build_runtime_profile(device), session=device.aws_session)
            for device in aws_devices
        ]

    @cache_results(ttl=120)
    def get_device(self, device_id: str) -> qbraid.runtime.braket.BraketDevice:
        """Returns the AWS device."""
        try:
            region_name = device_id.split(":")[3]
        except IndexError as err:
            raise ValueError(
                f"Device ARN is not a valid format: {device_id}. For valid Braket ARNs, "
                "see 'https://docs.aws.amazon.com/braket/latest/developerguide/braket-devices.html'"
            ) from err
        region_name = region_name or self._get_default_region()
        aws_session = self._get_aws_session(region_name=region_name)
        device = AwsDevice(arn=device_id, aws_session=aws_session)
        profile = self._build_runtime_profile(device)
        return BraketDevice(profile=profile, session=device.aws_session)
```

And to verify it works:

```python
from qbraid.runtime.braket import BraketProvider

import time

provider = BraketProvider()

start = time.time()

devices = provider.get_devices()

stop = time.time()

print(f"{stop - start:.2f} seconds")  # 12.59 seconds

start_cache = time.time()

devices_cached = provider.get_devices()

stop_cache = time.time()

print(f"{stop_cache - start_cache:.2f} seconds")  # 0.00 seconds
```

Give that a try

TheGupta2012:
Very cool implementation, thanks for the suggestion! 🔥 

One caveat was the cache access level. It was better to define cache on the provider instance level rather than global level to ensure - 

1. **There are no cache collisions** : Two providers defining the `get_devices` method in identical ways, `def get_devices(**kwargs)`, might lead to collision as `(func_name, *args, **kwargs)` (the tuple used for hashing) would be same

2. **Unit tests run easily** : When running unit tests, we want to update the return values of different methods inside tests. Even if we define new `QuantumProvider` instances, the global cache interfered with a later test.

ryanhill1 (changes requested):
Looks great! Like we discussed, would just like to add an some option that allows the user to bypass the cache and refetch the device data if they choose

TheGupta2012:
> Looks great! Like we discussed, would just like to add an some option that allows the user to bypass the cache and refetch the device data if they choose

Hey, I felt it was cleaner to let `bypass_cache` be a `kwarg`, handle it inside the decorator and not add it everywhere in the `get_devices` / `get_device` child implementations.

ryanhill1 (changes requested):
Can you please add the `bypass_cache` kwarg to each of the relevant function signatures for clarity? For example:

```python
@cache_results()
def get_devices(self, bypass_cache: bool = True, **kwargs):
    ...

@cache_results()
def get_device(self, device_id: str, bypass_cache: bool = True):
    ...
```

Otherwise, users would have to refer to the docstring to know that this option is even available.

Also, please update `CHANGELOG.md`. Thanks!

TheGupta2012:
Okay sure, I'll do that.

PR #771, opened 2024-10-02
Improve support for gates in `qasm2_to_ionq` conversion

### LINKED ISSUE #769: Add support for full register reference in `qasm2_to_ionq` converter
### Environment

- **qBraid version**: 0.8.2
- **Python version**: 3.12.5
- **Operating system**: macOS 14.4

### What happened?

The following code throws an error - 

```python3 
from qbraid.transpiler.conversions.qasm2 import qasm2_to_ionq

qasm_str = """
OPENQASM 2.0;\n qreg q[2];  creg c[2]; h q;
"""

print(qasm2_to_ionq(qasm_str))
```

```bash
  File "/Users/thegupta/Desktop/qBraid/sandbox/api-playground/qasm2_to_ionq.py", line 7, in <module>
    print(qasm2_to_ionq(qasm_str))
          ^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/thegupta/Desktop/qBraid/repos/qBraid/qbraid/transpiler/conversions/qasm2/qasm2_to_ionq.py", line 55, in qasm2_to_ionq
    _ = qubit.name.name
        ^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'name'
```

### Suggestions (Optional)

The `qasm2_to_ionq` converter does not take into account gate application on complete registers. Will need to extend that. Specifically [this](https://github.com/qBraid/qBraid/blob/b93399a647cd08789339a51a169bd8c07cdaace8/qbraid/transpiler/conversions/qasm2/qasm2_to_ionq.py#L54) processing assumes that the gate arguments are always single qubits which is not the case

TheGupta2012:
Fixes #769 
## Summary of changes
- Update the `qasm2_to_ionq` conversion to include full registers being applied as quantum gate arguments 
- Moreover, the number of qubits used in gate application is not reflective of the number of qubits on which a gate acts on. Eg. `h q[0], q[1];` means `h` is being applied on 2 qubits but is itself a 1-qubit gate. Updated the logic to identify this based on gate name instead of actual qubit arguments.
- Some other updates for representing the bits of qasm programs for easier handling

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -159,6 +159,30 @@ result = job.result()
 print(f"device_id: {result.device_id}, job_id: {result.job_id}, success: {result.success}")
 print(result.data.measurements)  # List of AhsShotResult instances
 ```
+- Improved the `qbraid.transpiler.conversions.qasm2.qasm2_to_ionq` method by adding support for registers in quantum gate arguments. See example below - ([#771](https://github.com/qBraid/qBraid/pull/771))
+
+```python
+from qbraid.transpiler.conversions.qasm2 import qasm2_to_ionq
+
+qasm_str = """
+OPENQASM 2.0;
+include "qelib1.inc";
+qreg q[3];
+x q[0];
+y q;
+"""
+
+print(qasm2_to_ionq(qasm_str))
+```
+
+outputs - 
+
+```python
+{'qubits': 3, 'circuit': [{'gate': 'x', 'target': 0}, 
+                          {'gate': 'y', 'target': 0}, 
+                          {'gate': 'y', 'target': 1}, 
+                          {'gate': 'y', 'target': 2}]}
+```
 
 ### Deprecated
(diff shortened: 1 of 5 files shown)


ryanhill1 (changes requested):
Looks great! Can you just add a quick update to the `CHANGELOG.md`

ryanhill1 (commented):
This reminds me of similar logic we have implemented here:

https://github.com/qBraid/qBraid/blob/96cde2676ec1cb5c53519b546172d0912de88148/qbraid/passes/qasm3/compat.py#L78-L123

Do you think that we could re-use or further generalize the `replace_gate_name()` function to help minimize redundancy?

TheGupta2012:
Well, in the ionq method we are creating a new representation i.e. the ionq json type and _not replacing the gate name in the qasm string_. 

Besides that, if we were to use the `replace_gate_name` method, it would have to be iterated over each gate key-value pair, which in turn would search through whole qasm string for each replacement ( amounting to an additional complexity of `O(|N|.|s|)` for `|N|` kv pairs and `|s|` length qasm string ). At the end, we do not need to process the updated qasm further so I feel the current approach would be a better fit for the ionq conversion.

ryanhill1 (approved):

PR #857, opened 2024-12-10
OpenQASM 3.0 to CUDA-Q Transpiler

### LINKED ISSUE #823: [FEATURE] Transpiler conversion function: OpenQASM to CUDA-Q
Would like to add a function to convert an `openqasm.ast.Program` to a `cudaq.kernel.kernel_builder.PyKernel`.

For example, would satisfy the following test case: 

```python
from __future__ import annotations

import textwrap
from typing import TYPE_CHECKING

import cudaq
import pytest
from openqasm3.parser import parse

from qbraid.transpiler.conversions.openqasm3 import openqasm3_to_cudaq

if TYPE_CHECKING:
    from cudaq.kernel.kernel_builder import PyKernel
    from openqasm3.ast import Program


@pytest.fixture
def cudaq_kernel() -> PyKernel:
    kernel = cudaq.make_kernel()

    qubit = kernel.qalloc()

    kernel.h(qubit)
    kernel.x(qubit)
    kernel.y(qubit)
    kernel.z(qubit)
    kernel.t(qubit)
    kernel.s(qubit)

    kernel.mz(qubit)

    return kernel


@pytest.fixture
def qasm_program() -> Program:
    qasm_str = """
    OPENQASM 3.0;
    include "stdgates.inc";

    qubit[1] q;
    bit[1] b;

    h q[0];
    x q[0];
    y q[0];
    z q[0];
    t q[0];
    s q[0];

    b[0] = measure q[0];
    """
    qasm_str = textwrap.dedent(qasm_str).strip()

    return parse(qasm_str)


def test_openqasm3_to_cudaq(cudaq_kernel, qasm_program):
    """Test converting an OpenQASM3 program to a CUDA-Q kernel."""
    cudaq_out = openqasm3_to_cudaq(qasm_program)
    assert str(cudaq_out) == str(cudaq_kernel)
```

arulandu:
## Summary of changes
OpenQASM 3.0 to CUDA-Q transpiler and unit tests. Closes #823.


arulandu:
At the moment, this draft doesn't support includes, custom gates, or many other OpenQASM 3.0 language features. As  @TheGupta2012 suggested, I think it's best if we pipe this through pyqasm for preprocessing so the transpiler can assume the program is well-formed (ex. rotation gates don't have more than one argument). This would also let pyqasm take care of includes and unfolding etc. 

Useful info I've found:
* `cudaq.register_operation` lets you register custom unitaries given a matrix custom operations 
* Gate modifiers in cudaq (specifically ctrl) works similar to OpenQASM. It would be nice if we could unroll everything, even CX, to just elementary gates + modifiers. Then, transpilation could be much simpler

arulandu:
The transpiler now support control/adjoint modifiers, but `cudaq.translate(_, format="openqasm2")` doesn't support control modifiers on gates and I've found a bug regarding the adjoint modifier:

image

Observe that while there's a `quake.apply<adj>` in the cudaq mlir, the qasm2 simple applies an s and t gate instead. 

We need to find a different testing solution for these cases? `cudaq.translate` to QIR works, but do we have a Qiskit -> QIR transpiler + QIR `allclose()`-like util for testing? Until then, the modifier tests fail

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+            if isinstance(inds, ast.DiscreteSet):
+                inds = inds.values
+
+            if len(inds) > 1:
+                raise ProgramConversionError(
+                    f"Discrete set or multi-integer indexing is unsupported: {inds}"
+                )
+
+            ind = inds[0]
+            if not isinstance(ind, ast.IntegerLiteral):
```
TheGupta2012, line 103: We don't need this validation, `pyqasm` handles the indexing for all indexed identifiers

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+        return q
+
+    for statement in program.statements:
+        if isinstance(statement, ast.Include):
+            if statement.filename == "stdgates.inc":
+                continue
+            raise ProgramConversionError(f"Includes are unsupported: {statement}")
+        if isinstance(statement, ast.QubitDeclaration):
+            ctx[statement.qubit.name] = kernel.qalloc(statement.size.value)
+        elif isinstance(statement, ast.ClassicalDeclaration):
```
TheGupta2012, line 119: I'm curious to know how cuda-q handles classical declarations with a type other than `bit`

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+            qubit_refs = [qubit_lookup(q) for q in qubits]
+
+
+            if len(statement.modifiers) > 1:
+                raise ProgramConversionError(
+                    f"Multiple gate modifiers are not supported: {statement}"
+                )
+            elif len(statement.modifiers) == 1:
+                mod = statement.modifiers[0]
+                if mod.modifier != ast.GateModifierName.ctrl:
```
TheGupta2012, line 156: This can be extended to use [the logic in `pyqasm`](https://github.com/qBraid/pyqasm/blob/main/src/pyqasm/visitor.py#L922) unroller. Essentially, we evaluate any `power` modifiers and see if we have an odd number of `inv` modifiers. I presume that if we're unrolling first then this would not be an issue as we won't propagate these modifiers here

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+if TYPE_CHECKING:
+    from cudaq import PyKernel, QuakeValue
+
+
+def gate_kernel(name: str, *args) -> PyKernel:
+    if name in ["x", "y", "z", "h", "s", "t"]:
+        size, argc = 1, 0
+    elif name in ["rx", "ry", "rz"]:
+        size, argc = 1, 1
+    elif name in ["swap"]:
```
TheGupta2012, line 35: `cx` / `cnot` gates are not supported?

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+
+"""
+Module containing OpenQASM to CUDA-Q conversion function
+
+"""
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Union
+
+import cudaq
```
ryanhill1, line 19: This will need to be imported lazily to avoid an `ImportError` in `qbraid/transpiler/conversions/__init__.py` if `cuda-quantum` is not installed.

Either by importing outside top level (inside function) or with `qbraid_core.LazyLoader`.

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+def identifier_lookup(cbit: Union[ast.IndexedIdentifier, ast.Identifier]) -> str:
+    """Unpacks the identifier name"""
+    if isinstance(cbit, ast.IndexedIdentifier):
+        return cbit.name.name
+
+    return cbit.name
+
+
+@weight(1)
+def openqasm3_to_cudaq(program: ast.Program) -> PyKernel:
```
ryanhill1, line 68: Need the `@requires_extras("cudaq")` decorator as well (imported from `qbraid.transpiler.annotations`)

This is so that in [`ConversionGraph.create_conversion_graph`](https://github.com/qBraid/qBraid/blob/896d6fb66da6436ee6796d36d17f5ea834eb9ed9/qbraid/transpiler/graph.py#L144-L146), the `openqasm3_to_cudaq` conversion is not included if [`Conversion.supported`](https://github.com/qBraid/qBraid/blob/896d6fb66da6436ee6796d36d17f5ea834eb9ed9/qbraid/transpiler/edge.py#L99) returns `False`, indicating that `cudaq` failed to import

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,172 @@
...
+            assert isinstance(qubit, ast.Identifier)
+            q = ctx[qubit.name][0]
+        return q
+
+    for statement in program.statements:
+        if isinstance(statement, ast.Include):
+            if statement.filename == "stdgates.inc":
+                continue
+            raise ProgramConversionError(f"Includes are unsupported: {statement}")
+        if isinstance(statement, ast.QubitDeclaration):
```
ryanhill1, line 117:
```suggestion
        if isinstance(statement, ast.Include):
            if statement.filename not in {"stdgates.inc", "qelib1.inc"}:
                raise ProgramConversionError(f"Unsupported include statement: '{statement.name}'")
        elif isinstance(statement, ast.QubitDeclaration):
```

tests/transpiler/qasm/test_qasm_to_cudaq.py
```diff
@@ -0,0 +1,159 @@
...
+#
+# THERE IS NO WARRANTY for the qBraid-SDK, as per Section 15 of the GPL v3.
+
+"""
+Unit tests for OpenQASM 3.0 to CUDA-Q kernel transpilation.
+
+"""
+from __future__ import annotations
+
+import cudaq
```
ryanhill1, line 17: skip tests if `cudaq` not installed

ryanhill1 (changes requested):

ryanhill1:
> ...
> * [ ]  Expressions? Parse and eval binary operators `rx(pi/4)` for example. `pyqasm` could help here?

@arulandu There's [`qbraid.passes.qasm.normalize_qasm_gate_params`](https://github.com/qBraid/qBraid/blob/896d6fb66da6436ee6796d36d17f5ea834eb9ed9/qbraid/passes/qasm/compat.py#L283) which converts `pi` symbols to decimal and then attempts to simplify any arithmetic expressions inside the parens. For example:

```python
from qbraid.passes.qasm import normalize_qasm_gate_params

qasm = """
OPENQASM 3;
qubit[1] q;
ry(3*pi/4) q[0];
"""

qasm_out = normalize_qasm_gate_params(qasm)

print(qasm_out)
```
```qasm
OPENQASM 3;
qubit[1] q;
ry(2.356194490192345) q[0];
```

And should also work for just the gate string as well e.g.

```python
gate = "ry(3*pi/4)"

gate_out = normalize_qasm_gate_params(gate)

print(gate_out)
```
```qasm
ry(2.356194490192345)
```

Should work for vast majority of cases. For more complicated expressions, e.g. that require "PEMDAS", it's a little buggy. Will open an issue for that.

ryanhill1:
> ...
> Thinking of using pyqasm `.unroll()` to handle some of the unfolding + expression eval?

@arulandu Yes, imo, this should be the first step of the function. For example:

```python
try:
    module = pyqasm.load(program)
    module.validate()
except Exception as err:
    raise ProgramConversionError("Program is not well-formed") from err

module.unroll()
program = module.unrolled_ast
```

@TheGupta2012 Question: For a `pyqasm.QasmModule` object, if it passes `module.validate()` without raising an exception, there's no need to include `module.unroll()` in the `try`/`except`, right?

TheGupta2012:
Yep, `unroll` should be okay out of the exception block.

ryanhill1:
> ...
> We need to find a different testing solution for these cases? `cudaq.translate` to QIR works, but do we have a Qiskit -> QIR transpiler + QIR `allclose()`-like util for testing? Until then, the modifier tests fail

@arulandu There's [`qbraid.transpiler.conversions.qiskit_to_pyqir`](https://github.com/qBraid/qBraid/blob/896d6fb66da6436ee6796d36d17f5ea834eb9ed9/qbraid/transpiler/conversions/qiskit/qiskit_extras.py#L54), but no `allclose` method for QIR that I know of.

Other options:
1. Investigate if there's a way to get the unitary matrix of a `cudaq.PyKernel`. Then we could start with a `qiskit` circuit, calculate it's unitary, convert `qiskit` $\rightarrow$ `qasm3` $\rightarrow$ `openqasm3` $\rightarrow$ `cudaq` and compare unitaries via [`assert_allclose_up_to_global_phase`](https://github.com/qBraid/qBraid/blob/896d6fb66da6436ee6796d36d17f5ea834eb9ed9/qbraid/interface/circuit_equality.py#L64)
2. (Probably best) Dig through the cuda-quantum repository for the `cudaq` test cases, and try to find any examples of how they compare the equivalence of two `cudaq.PyKernel` objects. I'm sure somewhere in there, they generate some `PyKernel` object and compare it to an "expected" value. In our test cases, could then just mirror that protocol.

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
arulandu: Will look into this further, but float is definitely allowed.

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
arulandu: Got it, will remove

arulandu:
> ```qasm
> ry(2.356194490192345)
> ```
> 
> This will work for the vast majority of cases.

sounds good. got this to work, but had to bump down the `allclose` precision to accommodate.

arulandu:
Pasting from Slack: this [issue](https://github.com/NVIDIA/cuda-quantum/issues/1822) says there's no current way to get a unitary from a kernel (though it's on the roadmap). But, we can find it by just calling `cudaq.get_state()` on a set of basis states to get every matrix element and then we just `np.allclose`. This suffices for our `random_circuit()` tests since they are low qubit count. For slightly higher # of qubits, we do this for randomized states, for higher qubit counts, this ofc is not feasible. 

Most `cudaq` tests look like [this](https://github.com/NVIDIA/cuda-quantum/blob/main/test/Translate/basic.qke) which seems to be string comparison. [This](https://github.com/NVIDIA/cuda-quantum/blob/main/test/Transforms/mapping_unitaries.qke) calls `Circuit-Check`, but I'm still trying to find how this test format works since everything is in comments.

arulandu:
**Findings from Testing**
* Transpiler fails for unary expressions like `gate g(t) q  { rx(-t) q; }` even after normalizing gate arguments 
* `pyqasm` unrolls most of the `stdgate` defs, but `cx` isn't unrolled to `ctrl @ x` likely because the `ctrl` modifier isn't supported yet. Once it is, perhaps we should always unroll gates to pure gates (non-control or adjoint) + modifiers

arulandu:
@ryanhill1 The method I mentioned previously to get the unitary only works if the kernel is parameterized by the `input_state`, that is the kernel takes in some `input_state` as a parameter and allocates `.qvector(input_state)`. However, the kernels generated by our transpiler don't do this because the OpenQASM programs don't take the state as a parameter either. Such kernels have no parameters. 

My workaround is **statevector testing**:
- `PyKernel` -> `cudaq.get_state()` -> amplitudes
- `qasm3` -> `QuantumCircuit` (qiskit) -> `StatevectorSimulator` -> amplitudes
- `np.allclose`

Transpiler now passes randomized testing for on Clifford gateset except (`"i", "sx", "sxdg", "ecr", "dcx", "iswap"`) with no measurements 🚀  Is it fine to use `qiskit-aer` as a dev-dependency for the statevector sim?

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,192 @@
...
+    if isinstance(cbit, ast.IndexedIdentifier):
+        return cbit.name.name
+
+    return cbit.name
+
+
+@weight(1)
+@requires_extras("cudaq")
+# pylint: disable-next=too-many-statements
+def openqasm3_to_cudaq(program: ast.Program) -> PyKernel:
```
TheGupta2012, line 78:
```suggestion
def openqasm3_to_cudaq(program: ast.Program | str) -> PyKernel:
```

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,192 @@
...
+    """
+    try:
+        module = pyqasm.load(program)
+        module.validate()
+    except Exception as e:
+        raise ProgramConversionError("PyQasm program is not well-formed") from e
+
+    module.unroll()
+    program = module.unrolled_ast
+    program = parser.parse(normalize_qasm_gate_params(printer.dumps(program)))
```
TheGupta2012, line 95: You don't need to add a call to `normalize_qasm_gate_params` as `unroll` operation by default substitutes the values of all variables inside the unrolled qasm.

tests/transpiler/qasm/test_qasm_to_cudaq.py
```diff
@@ -0,0 +1,329 @@
...
+    h q[0];
+    x q[0];
+    y q[0];
+    z q[0];
+    t q[0];
+    s q[0];
+
+    b[0] = measure q[0];
+    """
+    qasm3_in = parse(qasm3_str_in)
```
TheGupta2012, line 84: If we are allowing `str` in the `openqasm3_to_cudaq`, you can directly pass in the qasm string. 
Same comment applies for all tests.

ryanhill1 (changes requested):
Could you refactor `openqasm3_to_cudaq` to comply with the `pylint` warnings for `too-many-statements` and `too-many-nested-blocks`? Will help make the code more readable + maintainable

arulandu:
I could isolate functionality, but arguably it's a lot less clean to have to pass all the state around in arguments? I had ignored this because `openqasm3_to_ionq` ignores these linter statements as well

ryanhill1:
Fair enough we'll loop back later if necessary. Last thing would be to add tests for the ~24 missing lines

arulandu:
Current coverage is `94%   135, 163-168`. Line 135 is untestable because pyqasm doesn't support statements like `bit b = measure q` without throwing an error (even though I believe it should?). Line 163-168 are untestable until pyqasm adds support for ctrl modifiers.

arulandu:
@ryanhill1 Current coverage is `94%   135, 163-168`. Line 135 is untestable because pyqasm doesn't support statements like bit b = measure q without throwing an error (even though I believe it should?). Line 163-168 are untestable until pyqasm adds support for ctrl modifiers.

TheGupta2012:
@arulandu You're correct, pyqasm does not yet support the `ctrl` modifiers but it does support the `b = measure q;` type statements - 

```python
In [8]: import pyqasm

In [9]: qasm_str = """
   ...: OPENQASM 3;
   ...: qubit[4] q;
   ...: bit[4] result;
   ...: result = measure q;
   ...: """

In [10]: mod = pyqasm.loads(qasm_str)

In [11]: mod.unroll()

In [12]: pyqasm.dumps(mod).splitlines()
Out[12]: 
['OPENQASM 3.0;',
 'qubit[4] q;',
 'bit[4] result;',
 'result[0] = measure q[0];',
 'result[1] = measure q[1];',
 'result[2] = measure q[2];',
 'result[3] = measure q[3];']

In [13]: pyqasm.__version__
Out[13]: '0.1.0'
```

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+            if statement.target is not None:
+                # pyqasm unrolls all identifiers to indexed identifiers
+                assert isinstance(statement.target, ast.IndexedIdentifier)
+                ctx[statement.target.name.name] = val
+        elif isinstance(statement, ast.QuantumGate):
+            name, qubits = statement.name.name, statement.qubits
+
+            args = []
+            for arg in statement.arguments:
+                # pyqasm unrolls imples gate arguments are literals
```
TheGupta2012, line 151:
```suggestion
                # pyqasm unrolls implies gate arguments are literals
```

arulandu:
@TheGupta2012 yes, but I don't think it supports `bit[4] result = measure q'` in one statement?

ryanhill1:
@TheGupta2012  is `bit[4] result = measure q` valid syntax in OpenQASM 3? Specifically `bit[4]` instead of just `bit`

ryanhill1 (changes requested):
Can we replace some of the `assert` statements with a `ValueError` + error message where it makes sense? Hard for a user to debug when it's just an assertion error. The comments you provide above the assertion errors are great, but a user may not see them unless they really dig. Not relevant for all of them, but definitely for some e.g. anything not currently supported by pyqasm

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+        "rz",
+        "h",
+        "s",
+        "t",
+        "sdg",
+        "tdg",
+        "u3",
+        "i",
+        "id",
+        "iden",
```
ryanhill1, line 52: I think if we have `i` and `id` that's good. Not sure `iden` is as common of a convention, unless that's used by cudaq. lmk

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+        module = pyqasm.loads(program)
+        module.validate()
+    except Exception as e:
+        raise ProgramConversionError("QASM program is not well-formed.") from e
+
+    module.unroll()
+    program = module.unrolled_ast
+
+    kernel: PyKernel = cudaq.make_kernel()
+    ctx: Dict[str, Union[QuakeValue | None]] = {}
```
ryanhill1, line 98:
```suggestion
    ctx: dict[str, Optional[QuakeValue]] = {}
```

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+    gate_kernels: Dict[str, PyKernel] = {}
```
ryanhill1, line 99:
```suggestion
    gate_kernels: dict[str, PyKernel] = {}
```

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def get_gate(name: str, targs: tuple[type]) -> PyKernel:
+        if name in gate_kernels:
+            return gate_kernels[name]
+
+        gate_kernels[name] = make_gate_kernel(name, targs)
+        return gate_kernels[name]
+
+    def qubit_lookup(qubit: Union[ast.IndexedIdentifier, ast.Identifier]) -> QuakeValue:
```
ryanhill1, line 108:
```suggestion
    def qubit_lookup(qubit: ast.IndexedIdentifier | ast.Identifier) -> QuakeValue:
```

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+    else:
+        op(*qrefs)
+
+    return kernel
+
+
+@weight(1)
+@requires_extras("cudaq")
+# pylint: disable-next=too-many-statements
+def openqasm3_to_cudaq(program: Union[QasmStringType, ast.Program]) -> PyKernel:
```
ryanhill1, line 79:
```suggestion
def openqasm3_to_cudaq(program: QasmStringType | ast.Program) -> PyKernel:
```

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
ryanhill1, line 1: Added a few typing update suggestions as well because now that we've recently updated the project to Python >= 3.10

ryanhill1 (changes requested):

TheGupta2012:
> @TheGupta2012 In OpenQASM 3, can you declare a bit and measure it on the same line? e.g.
> 
> ```qasm
> OPENQASM 3;
> qubit q;
> bit b = measure q;
> ```
> 
> `pyqasm` doesn't like it, but not sure if it's valid or not within the OpenQASM 3 language spec

There isn't any mention that it's not compatible. Looking at the grammar for [openqasm classical declaration](https://openqasm.com/grammar/index.html), I feel it should be allowed. Will open up an issue!

ryanhill1:
@arulandu Anything left to do here or is this PR now up-to-date with all gates supported by cuda-q?

arulandu:
@ryanhill1 I think everything that cudaq has natively, we should already support. If we don't support something, it's blocked by a pyqasm issue usually (ex. multi-bit ctrl)

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Types of changes:
 ### Fixed
 
 ### Added
+- Added support for OpenQASM 3.0 to CUDA-Q kernel transpilation ([#857](https://github.com/qBraid/qBraid/pull/857))
```
TheGupta2012, line 20: Can you add a usage example here?

qbraid/transpiler/conversions/openqasm3/openqasm3_to_cudaq.py
```diff
@@ -0,0 +1,188 @@
...
+                args.append(arg.value)
+            targs = [type(a) for a in args]
+
+            qubit_refs = [qubit_lookup(q) for q in qubits]
+
+            # pyqasm unrolls multiple modifiers.
+            # ctrl isn't supported so multi-ctrl is not an issue at the moment.
+            assert len(statement.modifiers) < 1
+
+            if len(statement.modifiers) == 1:
```
TheGupta2012, line 161: Would this case occur if we're asserting the length to be < 1? Or maybe the assertion is supposed to be for `len(statement.modifiers) <= 1`?

ryanhill1 (approved):

PR #883, opened 2025-01-27
program serialization and load_job

ryanhill1:
## Summary of changes

- Added `qbraid.runtime.load_job` function that uses entrypoints to load provider job class and create instance with job id
- Switched all `QbraidJob` sub-classes to only require `job_id` as positional argument, and any other args that used to be required for auth can now be loaded with credentials from environment variables 
- Added `QuantumProgram.serialize` method to streamline creation of `ProgramSpec` classes in `QbraidProvider`


cosenal:
Interesting, I had just made [this change](https://github.com/unitaryfund/metriq-gym/pull/174/files#diff-20a738b0fa0da652a8b9290ca8f67375b9d95506a70907d0248c8f34e7bd0d67R35-R58) in our metriq-gym project, because IonQ jobs were requiring session to be provided. Probably that's no longer needed after your change.

Issue #848, opened 2024-12-06
Move `circuit_drawer` function to abstract `draw` method in `GateModelProgram`


ryanhill1:
The [circuit_drawer](https://github.com/qBraid/qBraid/blob/ffcdc383dd7f038c5cd79aeed262b17fe3783bcc/qbraid/visualization/draw_circuit.py#L29) function currently uses `get_program_type_alias` to get the input quantum program type, and based on what package it is, imports and invokes the necessary draw functionality using that package.

Instead, we should add an abstract `draw` method to the [GateModelProgram](https://github.com/qBraid/qBraid/blob/ffcdc383dd7f038c5cd79aeed262b17fe3783bcc/qbraid/programs/gate_model/_model.py#L28) class, and implement the draw method for each package in [qbraid.programs.gate_model](https://github.com/qBraid/qBraid/tree/ffcdc383dd7f038c5cd79aeed262b17fe3783bcc/qbraid/programs/gate_model). The available keyword arguments should be defined based on those available for that package. The current `output` and `as_package` arguments of the current `circuit_drawer` method will therefore not be needed.

PR #916, opened 2025-03-06
Solving the first issue

pencil-dev:


ryanhill1:
Hi @pencil-dev, could you reference which issue this is solving and summarize the changes?

ryanhill1 (changes requested):
If you're aiming to resolve both https://github.com/qBraid/qBraid/issues/848 and https://github.com/qBraid/qBraid/issues/848, do you min breaking it into two separate PRs?

PR #947, opened 2025-03-31
Azure runtime - Pasqal provider + job/result handling

rryoung98:
Update the copyright year to 2025 and improve Azure job handling by adding a method to check device availability and modifying job submission to include a target specification parameter. This enhances the overall functionality and clarity of the Azure integration.


rryoung98:
@ryanhill1  I'm going to really polish this PR, as it still needs more testing, documentation updates etc, but I'm going to make it "Ready for review" since it's been switched to use the preferred `_serialize` method

qbraid/runtime/azure/job.py
```diff
@@ -96,7 +99,7 @@ def _make_estimator_result(data: dict[str, Any]) -> MicrosoftEstimatorResult:
         result_data = data["data"]
         return MicrosoftEstimatorResult(result_data)
 
-    def result(self) -> Union[Result, MicrosoftEstimatorResult]:
+    def result(self, wait_until_completed=False) -> Union[Result, MicrosoftEstimatorResult]:
```
ryanhill1, line 102: Why this? If job is not completed, result will not contain any data.

qbraid/runtime/azure/result_builder.py
```diff
@@ -346,3 +346,66 @@ def get_counts(self) -> Union[dict[str, int], list[dict[str, int]]]:
             return results[0]["data"]["counts"] if results[0]["success"] else {}
 
         return [result["data"]["counts"] if result["success"] else {} for result in results]
+
+
+class AzureAHSModelResultBuilder:
```
ryanhill1, line 351: wouldn't it be easier / require less redundant code to just add a format ahs result method to the existing result builder class? seems to be a fair amount of overlap

ryanhill1 (changes requested):

qbraid/runtime/azure/result_builder.py
rryoung98: That's true, but it didn't make sense since from a naming convention to add an analog hamiltonian simulation formatter in a GateModelResultBuilder.

qbraid/runtime/azure/job.py
rryoung98: I noticed that wait_until_completed() was being called in job.py (prior to my edits), but its purpose wasn’t entirely clear. Since waiting for completion can introduce significant delays—especially with our API potentially taking a long time to return results—I decided to make this behavior configurable via a boolean flag. 

Reference:

[job.py line 109](https://github.com/qBraid/qBraid/blob/99b8bb33b226576f599c7ad22304dda11603a073/qbraid/runtime/azure/job.py#L109)

qbraid/runtime/azure/result_builder.py
rryoung98: If you're okay with the name change to AzureResultBuilder I can do that, just wasn't sure if the convention was to separate Gate vs AHS vs Annealing etc per provider similar to Braket.

qbraid/runtime/result_data.py
```diff
@@ -345,6 +344,15 @@ def __eq__(self, other):
 
         return all(s1 == s2 for s1, s2 in zip(self._measurements, other._measurements))
 
+    def __repr__(self) -> str:
+        """Return a string representation of the AhsResultData instance."""
+        measurements_info = (
+            "[" + ", ".join(repr(shot) for shot in self._measurements) + "]"
+            if self._measurements is not None
+            else "None"
+        )
```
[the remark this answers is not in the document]
ryanhill1: oh I guess maybe but I think we can just leave that for now and maybe open a separate issue for it

PR #949, opened 2025-04-04
runtime: Normalize Braket batch job result using BatchQuantumJob

### LINKED ISSUE #879: [FEATURE] Abstraction for batch job across providers
### Feature Description

Cloud providers have their own idiosyncrasies when it comes to handling batch jobs. 
qBraid's `device.run(circuits: list[QPROGRAM])` allows to run a list/batch of quantum tasks/circuits, but this gets handled differently by each cloud provider.
For example, consider IBM Qiskit vs. AWS Braket. Qiskit has an actual implementation of a batch job object that collects all the quantum tasks, whereas in Braket, each task is a job on its own. From the p.o.v of a qBraid-SDK user, when a batch of tasks is run on a device from `QiskitRuntimeProvider`, then a single `QuantumJob` is created, whose "result.get_counts()" will be a list of measurement counts. Instead, when the same batch of tasks is run on a device from `BraketProvider`, a list of `QuantumJob` is returned, and applying a "result.get_counts()" to each item of this list will yield a single measurement count (`mc`) object. In short,

### IBM
device.run([circuit_1, circuit_2]) --> job: QuantumJob --> job.result.get_counts() --> [mc_1, mc_2]

### AWS
device.run([circuit_1, circuit_2]) --> [job_1, job_2]: list[QuantumJob] --> [job_1.result.get_counts(), job_2.result.get_counts()] --> [mc_1, mc_2]

When writing code in qBraid that is supposed to be agnostic of the provider, this becomes hairy to handle. It would be cool for the SDK users if qBraid tried to normalize those differences in how the different providers handle batch jobs.

### Implementation (Optional)

Without handling all the scenarios, we still make it work in metriq-gym at the moment, with code like this https://github.com/unitaryfund/metriq-gym/blob/2305e4a3b89b9d29d506c3852fa3e9a79c6cf9cf/metriq_gym/run.py#L74-L75

soheil-star01:
### Summary

This PR introduces a `BatchQuantumJob` wrapper to normalize the output of `BraketDevice.run()` when executing a batch of circuits. Previously, Braket returned a `list[QuantumJob]`, whereas providers like IBM returned a single `QuantumJob` object for batch runs. This discrepancy made provider-agnostic result handling more complex.

### What’s Changed
- Updated `BraketDevice.submit()` to return a `BatchQuantumJob` for batch submissions.
- Added a `BatchQuantumJob` class to aggregate multiple `QuantumJob` instances and unify the `.result()` API.
- Refactored `test_batch_run` to assert `BatchQuantumJob` is returned and added mocks for multi-task result handling.

### Closes
Fixes #879

### DIFF
--- a/qbraid/runtime/aws/batch_job.py
+++ b/qbraid/runtime/aws/batch_job.py
@@ -1,3 +1,13 @@
+# Copyright (C) 2024 qBraid
+#
+# This file is part of the qBraid-SDK
+#
+# The qBraid-SDK is free software released under the GNU General Public License v3
+# or later. You can redistribute and/or modify it under the terms of the GPL v3.
+# See the LICENSE file in the project root or <https://www.gnu.org/licenses/gpl-3.0.html>.
+#
+# THERE IS NO WARRANTY for the qBraid-SDK, as per Section 15 of the GPL v3.
+
 """
 Defines the BatchQuantumJob class used for grouping and managing multiple
 Braket quantum tasks submitted as a batch.


qbraid/runtime/aws/batch_job.py
ryanhill1, line 1: Let's move this file to `qbraid/runtime/batch_job.py` since it will eventually apply to all providers except IBM. For now, it's fine to keep the integration limited to AWS—we just want the file in a shared location for future use.

qbraid/runtime/aws/batch_job.py
```diff
@@ -0,0 +1,52 @@
...
+    """
+
+    def __init__(self, jobs: list[QuantumJob]):
+        super().__init__(job_id="batch", device=jobs[0].device)
+        self._jobs = jobs
+
+    def status(self):
+        statuses = [job.status() for job in self._jobs]
+        if all(s == JobStatus.COMPLETED for s in statuses):
+            return JobStatus.COMPLETED
```
ryanhill1, line 42: If all jobs have the same status, just return that status e.g.

```python
unique_statuses = set(statuses)
if len(unique_statuses) == 1:
    return unique_statuses.pop()
```

or

```python
if len(set(statuses)) == 1:
    return statuses[0]
```

qbraid/runtime/aws/batch_job.py
```diff
@@ -0,0 +1,52 @@
...
+This abstraction allows users to interact with multiple quantum jobs as a
+single job-like object, providing consistent access to results, metadata,
+status checking, and job cancellation across different quantum providers.
+"""
+
+from qbraid.runtime.enums import JobStatus
+from qbraid.runtime.job import QuantumJob
+
+
+class BatchQuantumJob(QuantumJob):
```
ryanhill1, line 24: can we call this a `QuantumJobBatch` instead?

qbraid/runtime/aws/device.py
ryanhill1, line 1: Perhaps for AWS, we should use their built-in [AwsQuantumTaskBatch](https://github.com/amazon-braket/amazon-braket-sdk-python/blob/v1.91.0/src/braket/aws/aws_quantum_task_batch.py#L42) abstraction, which didn't exist at the time this module was created.

qbraid/runtime/aws/batch_job.py
```diff
@@ -0,0 +1,52 @@
...
+    Used to normalize the return type of device.run(batch) across providers.
+    Provides a unified interface for checking status, collecting results,
+    and managing metadata of the entire batch as one logical job.
+
+    Attributes:
+        jobs (list[QuantumJob]): The individual quantum jobs in the batch.
+    """
+
+    def __init__(self, jobs: list[QuantumJob]):
+        super().__init__(job_id="batch", device=jobs[0].device)
```
ryanhill1, line 36: the `jobs[0].device` call might raise a `ResourceNotFoundError`. In that case we would just want to pass `device=None` instead of letting that exception be raised. We also probably want to require that all of the jobs were executed on the same device, and raise an exception if not

ryanhill1 (changes requested):

qbraid/runtime/aws/batch_job.py
soheil-star01: I originally made the logic explicit to preserve readability and express priority (e.g., COMPLETE > FAILED > RUNNING). While using set(statuses) would reduce lines, it may reduce clarity and doesn’t support the current explicit ordering. That said, I’m happy to refactor if you prefer more generic logic.

ryanhill1: I think the concern is that if all jobs in the batch share the same non-terminal status (e.g. `QUEUED`), labeling the batch status as `RUNNING` might be misleading. It makes more sense for the batch to inherit that shared status in such cases.

More generally:
- If all jobs share the same status, the batch should reflect that status.
- If there's a mix of non-terminal statuses, or a mix of non-terminal and terminal statuses, default to `RUNNING`.
- If all statuses are terminal and **any** job is `FAILED`, the batch should be `FAILED`.
- If all are terminal and `CANCELLED`, set the batch to `CANCELLED`.
- Only set the batch to `COMPLETED` if all jobs are completed

Does that logic sound right to you?

PR #956, opened 2025-04-21
Fix simulator check for Azure targets

cosenal:
## Summary of changes
Non-simulator target profiles on Azure follow the pattern `provider_id.qpu.device_id`, see https://learn.microsoft.com/en-us/azure/quantum/qc-target-list
Old code was not working as it'd compare the string `qpu.device_id` against `qpu`, producing ill results. 

### Smoke test
```python3
from qbraid.runtime import AzureQuantumProvider
provider = AzureQuantumProvider()
device = provider.get_device("ionq.simulator")
assert(device.simulator)
device = provider.get_device("ionq.qpu.aria-1")
assert(not device.simulator)
```

### Unit test
`tox -e unit-tests -- tests/runtime/azure/test_azure_runtime.py`


ryanhill1 (approved):
LGTM, thanks!

Yes, other test failures due to pyqasm bug in latest release: https://github.com/qBraid/pyqasm/issues/175

Patching with https://github.com/qBraid/qBraid/pull/957

Issue #704, opened 2024-07-07
Integrate `mypy`


ryanhill1:
1. Make qBraid-SDK repository pass checks from `mypy` static type checker
2. Integrate `mypy` into qBraid-SDK `tox` linters environment workflows

PR #986, opened 2025-05-28
Circumvent constructing unregistered program type nodes in `ConversionGraph`

### LINKED ISSUE #789: [BUG] "unregistered" program type still included in `ConversionGraph`
After calling `unregister_program_type("qasm2")`, the program type `"qasm2"` is removed from `QPROGRAM_REGISTRY`, but it is still included in the `ConversionGraph`. This leads to inconsistent behavior where `"qasm2"` is no longer recognized in the registry but still exists as a node in a new instance of `ConversionGraph`.

To reproduce:

```python
from qbraid import QPROGRAM_REGISTRY, ConversionGraph, unregister_program_type

print("qasm2" in QPROGRAM_REGISTRY)  # True

graph = ConversionGraph()

print(graph.has_node("qasm2"))  # True

unregister_program_type("qasm2")

print("qasm2" in QPROGRAM_REGISTRY)  # False

new_graph = ConversionGraph()

print(new_graph.has_node("qasm2"))  # True, but should be False
```

After unregistering `"qasm2"`, it should not be present in new instances of `ConversionGraph`. It should be a requirement that all "nodes" in the `ConversionGraph` are "registered" program types.

*Note*:  The "qasm2" alias was just used as an example. This issue is applicable to all program types. The solution you propose should generalize to all program types.

SaashaJoshi:
## Summary of changes

Okay, so I could figure out that the main error comes from loading the default conversions in this line,
https://github.com/qBraid/qBraid/blob/d3f6614b5ea3c58fe0fe1a0a79be68638673e26c/qbraid/transpiler/graph.py#L132-L147

This PR makes changes to only consider `source` and `target` in `conversion_functions` that are present/registered in `QPROGRAM_ALIASES`, hence solving the bug mentioned in #789. 

Closes #789.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ Types of changes:
 - Fixed lazy importing bug in `plot_histogram` method ([#972](https://github.com/qBraid/qBraid/pull/972))
 - Fixed bug which caused all `braket` conversions to be unavailable if `cirq` was not installed due to an eager top-level import in `braket_to_cirq.py` which should have been done lazily ([#982](https://github.com/qBraid/qBraid/pull/982))
 - Made Pulser unit test version-agnostic to support any installed Pulser version. ([#983](https://github.com/qBraid/qBraid/pull/983))
+- Fixed the bug that included unregistered program type in `ConversionGraph`. ([#986](SaashaJoshi:unregister))
 
 ### Dependencies
 - Migrated to `setuptools>=77` due to TOML-table based `project.license` deprecation in favor of SPDX expression in compliance with [PEP 639](https://peps.python.org/pep-0639) ([#973](https://github.com/qBraid/qBraid/pull/973))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,7 +32,7 @@ Types of changes:
 - Fixed lazy importing bug in `plot_histogram` method ([#972](https://github.com/qBraid/qBraid/pull/972))
 - Fixed bug which caused all `braket` conversions to be unavailable if `cirq` was not installed due to an eager top-level import in `braket_to_cirq.py` which should have been done lazily ([#982](https://github.com/qBraid/qBraid/pull/982))
 - Made Pulser unit test version-agnostic to support any installed Pulser version. ([#983](https://github.com/qBraid/qBraid/pull/983))
-- Fixed the bug that included unregistered program type in `ConversionGraph`. ([#986](SaashaJoshi:unregister))
+- Fixed the bug that included unregistered program type in `ConversionGraph`. ([#986](https://github.com/qBraid/qBraid/pull/986))
 
 ### Dependencies
 - Migrated to `setuptools>=77` due to TOML-table based `project.license` deprecation in favor of SPDX expression in compliance with [PEP 639](https://peps.python.org/pep-0639) ([#973](https://github.com/qBraid/qBraid/pull/973))
--- a/qbraid/transpiler/graph.py
+++ b/qbraid/transpiler/graph.py
@@ -138,19 +138,19 @@ class ConversionGraph(rx.PyDiGraph):
         """
         transpiler = import_module("qbraid.transpiler.conversions")
         conversion_functions: list[str] = getattr(transpiler, "conversion_functions", [])
-        conversion_functions = [conversion.split("_to_") for conversion in conversion_functions]
-        conversion_functions = [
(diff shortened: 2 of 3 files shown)


SaashaJoshi:
Hi @ryanhill1,

I observe that when doing the following,

```python
unregister_program_type("qasm2")

new_graph = ConversionGraph()
print(new_graph.has_node("qasm2"))  # Shows False, should be False.

new_graph_2 = ConversionGraph(nodes=["qasm2"])
print(new_graph_2.has_node("qasm2"))    # Shows True (Is this correct?).
```

Could you confirm if this behavior is correct? If not, I think the issue potentially stems from not initiating `nodes` correctly in the `ConversionGraph` class. Consider this PR WIP in this case.

CHANGELOG.md
```diff
@@ -32,6 +32,7 @@ Types of changes:
 - Fixed lazy importing bug in `plot_histogram` method ([#972](https://github.com/qBraid/qBraid/pull/972))
 - Fixed bug which caused all `braket` conversions to be unavailable if `cirq` was not installed due to an eager top-level import in `braket_to_cirq.py` which should have been done lazily ([#982](https://github.com/qBraid/qBraid/pull/982))
 - Made Pulser unit test version-agnostic to support any installed Pulser version. ([#983](https://github.com/qBraid/qBraid/pull/983))
+- Fixed the bug that included unregistered program type in `ConversionGraph`. ([#986](SaashaJoshi:unregister))
```
ryanhill1, line 35:
```suggestion
- Fixed the bug that included unregistered program type in `ConversionGraph`. ([#986](https://github.com/qBraid/qBraid/pull/986))
```

qbraid/transpiler/graph.py
```diff
@@ -138,10 +138,16 @@ def load_default_conversions(bias: Optional[float] = None) -> list[Conversion]:
         """
         transpiler = import_module("qbraid.transpiler.conversions")
         conversion_functions: list[str] = getattr(transpiler, "conversion_functions", [])
+        conversion_functions = [conversion.split("_to_") for conversion in conversion_functions]
+        conversion_functions = [
```
ryanhill1, line 142: Could you rename the `conversion_functions` variables on lines 141 and 142 to something different for clarity? Since the variable is explicitly typed as `list[str]` on line 140 and then changes type on subsequent lines, using distinct names would make the code easier to read.

Giving them distinct names will also help make the static type checkers happy (even though we're not strictly enforcing type checking right now, we are eventually planning to: https://github.com/qBraid/qBraid/issues/704)

ryanhill1 (changes requested):
These changes look good so far!

You raised a good point when you referenced:

> new_graph_2 = ConversionGraph(nodes=["qasm2"])
> print(new_graph_2.has_node("qasm2"))    # Shows True (Is this correct?).

I don't think one should be allowed to specify a node to use in the `ConversionGraph` that corresponds to program type that isn't registered. So ideally, I think that first line should raise an exception (maybe a [PackageValueError](https://github.com/qBraid/qBraid/blob/d3f6614b5ea3c58fe0fe1a0a79be68638673e26c/qbraid/programs/exceptions.py#L32)?).

qbraid/transpiler/graph.py
SaashaJoshi: Good catch!

tests/transpiler/test_conversion_graph.py
```diff
@@ -195,6 +195,28 @@ def test_copy_conversion_graph():
     assert node_alias_id_map_init == copy._node_alias_id_map
 
 
+@pytest.mark.parametrize("program_type", ["qasm2", "qasm3", "pyquil", "braket", "cirq"])
+def test_unregistered_node_in_conversion_graph(program_type):
+    """Test the unregistered nodes in ConversionGraph"""
+    assert program_type in QPROGRAM_ALIASES
```
ryanhill1, line 201:
```suggestion
    if program_type not in QPROGRAM_ALIASES:
        pytest.skip(f"{program_type} not installed")
```

`pyquil` is not compatible with Python 3.13, so is not installed during those unit-tests:

https://github.com/qBraid/qBraid/blob/d3f6614b5ea3c58fe0fe1a0a79be68638673e26c/requirements-dev.txt#L6

The change suggested above should fix it.

ryanhill1 (changes requested):

ryanhill1 (approved):

Issue #786, opened 2024-10-10
[BUG] `IonQJob.result()` for ideal simulator fails bc shots=None


ryanhill1:
Jobs run on the IonQ "ideal" simulator provide deterministic results, so for these jobs, the IonQ API returns results as histogram probabilities instead of measurement counts. However, currently, there is no way to pass just raw probability data to `GateModelResultData` (https://github.com/qBraid/qBraid/issues/785). To patch this, we've added an arbitrary default shots value to `IonQ.submit`:

https://github.com/qBraid/qBraid/blob/03aa38b89081a3e6fc6890ac4e785eba96769847/qbraid/runtime/ionq/device.py#L75

so that the `IonQJob.result()` method can construct measurement counts data based on the returned probabilities together with that shots value:

https://github.com/qBraid/qBraid/blob/03aa38b89081a3e6fc6890ac4e785eba96769847/qbraid/runtime/ionq/job.py#L102-L108

For sequential job submission and results retrieval this works fine:

```python
from qbraid.runtime import IonQJob, IonQProvider

provider = IonQProvider(api_key="...")

device = provider.get_device("simulator")

program = """
OPENQASM 2.0;
qreg q[2];
h q[0];
cx q[0],q[1];
cx q[1],q[0];
"""

job = device.run(program)

# Getting the result directly from the job object created from device.run
# works fine because we carry over arbitrary placeholder value for shots
print(job._cache_metadata.get("shots")) # 100

result = job.result()
```

However, this arbitrary shots value is not used by the IonQ "ideal" simulator, and therefore is not stored anywhere other than locally within that initial `IonQJob` instance. So if you then create a new job instance from the job ID, the `result()` method fails with a `ValueError` because there is no stored value for "shots."

```python
job = IonQJob(job.id, session=provider.session)

print(job._cache_metadata.get("shots")) # None

result = job.result()
# File ~/.../qBraid/qbraid/runtime/ionq/job.py:86, in IonQJob._get_counts(result)
#      84 probs_dec_str: Optional[dict[str, float]] = result.get("probabilities")
#      85 if shots is None or probs_dec_str is None:
# ---> 86     raise ValueError("Missing shots or probabilities in result data.")
#      87 probs_dec = {int(key): value for key, value in probs_dec_str.items()}
#      88 probs_normal = normalize_counts(probs_dec)

# ValueError: Missing shots or probabilities in result data.
```

## TODO

- Close related issue https://github.com/qBraid/qBraid/issues/785
[...]

PR #989, opened 2025-06-01
support GateModelResultData class with precomputed probability

### LINKED ISSUE #785: Enhance `GateModelResultData.get_probabilities()` to support direct probability data
The [`GateModelResultData.get_probabilities()`](https://github.com/qBraid/qBraid/blob/03aa38b89081a3e6fc6890ac4e785eba96769847/qbraid/runtime/result.py#L142) method currently computes a probability distribution based on the measurement counts:

https://github.com/qBraid/qBraid/blob/03aa38b89081a3e6fc6890ac4e785eba96769847/qbraid/runtime/result.py#L163-L164

This works well for QPUs and for simulators that return raw measurement counts. However, many **ideal quantum simulators** compute the exact quantum state, providing deterministic results. These simulators do not involve probabilistic sampling, and the concept of `shots` does not apply. Consequently, these simulators do not return raw measurements or measurement counts, but instead provide **direct probability data** (histogram).

To better support these simulators, the `GateModelResultData` class should be updated to optionally accept **pre-computed probabilities**. This would allow `get_probabilities()` to return these probabilities directly, without needing to calculate them from counts.

### TODO

 - Add an optional argument to the `GateModelResultData` class to accept **pre-computed probabilities**.
 - If probabilities are provided, `get_probabilities()` should return them directly.
 - If probabilities are not provided, `get_probabilities()` should maintain its current behavior of calculating probabilities based on `get_counts()`.
 - Regardless of whether probabilities are computed from counts or provided directly, `get_probabilities()` should continue to respect the `decimal=True/False` option for formatting keys.
 - Since the function for normalizing data will now apply to both counts and direct probability histograms, consider renaming [`normalize_counts()`](https://github.com/qBraid/qBraid/blob/03aa38b89081a3e6fc6890ac4e785eba96769847/qbraid/runtime/postprocess.py#L157) to something more general, like `normalize_histogram_data()`.

Hazarre:
Closes #785

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ Types of changes:
 ### Improved / Modified
 - Updated `TimeStamps` schema to auto-compute `executionDuration` from `createdAt` and `endedAt` if not explicitly provided. ([#983](https://github.com/qBraid/qBraid/pull/983))
 - Enhanced `TimeStamps` to accept both `datetime.datetime` objects for `createdAt` and `endedAt` (previously only accepted ISO-formatted strings). ([#983](https://github.com/qBraid/qBraid/pull/983))
+- Enhance `GateModelResultData` to support direct/pre-computed probability data ([#785](https://github.com/qBraid/qBraid/issues/785))
 
 ### Deprecated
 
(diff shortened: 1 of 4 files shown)


ryanhill1 (commented):
Thanks for this PR! Unfortunately I've chose to merge PR https://github.com/qBraid/qBraid/pull/988 to close this issue, since it was a little bit more complete. But we would love to have you as a contributor if you find another issue that you are interested in! One issue that could be very well suited as a follow-up to this one is https://github.com/qBraid/qBraid/issues/786. Would you be interested in working on that?

PR #990, opened 2025-06-02
Fix a `figsize` argument for graph plots.

### LINKED ISSUE #851: [BUG] Fix nodes being cut-off around outside of conv graph plot
In [`qbraid.visualization.plot_conversion_graph()`](https://github.com/qBraid/qBraid/blob/693890df2599c36d66ce3f73d8e156f75a30b131/qbraid/visualization/plot_conversions.py#L64), certain seeds can cause nodes to be partially or entirely cut off near the edges of the graph. For instance, in the example below, the "qubo" node in the top-left corner and the "cpp_pyqubo" node label in the bottom-right corner are clipped. To address this issue, we propose enhancing `plot_conversion_graph` to ensure that all nodes and their labels are fully visible within the plot boundaries. One solution could be some additional parameters passed to [`rx.spring_layout`](https://github.com/qBraid/qBraid/blob/693890df2599c36d66ce3f73d8e156f75a30b131/qbraid/visualization/plot_conversions.py#L146), which determines the placement of nodes in the graph layout. Other implementations/approaches are also possible.

```python
from qbraid import ConversionGraph

graph = ConversionGraph(include_isolated=True)

graph.plot(seed=362, legend=True)
```

52f4331a-9cb3-4073-b109-1b31a7d32286

SaashaJoshi:
## Summary of changes
This PR fixes a `figsize` argument for the matplotlib graph plots such that the nodes are not cropped off from the image edge while saving. Potentially closes #851

### DIFF
--- a/qbraid/visualization/plot_conversions.py
+++ b/qbraid/visualization/plot_conversions.py
@@ -158,6 +158,7 @@ def plot_conversion_graph(  # pylint: disable=too-many-arguments
         kwargs["edgecolors"] = edgecolors
 
     plt.ioff()  # Disable interactive mode
+    plt.figure(layout="constrained", figsize=(8, 6))
 
     mpl_draw(
         graph,


SaashaJoshi:
This is a sample `.png` image that gets saved on my system.

png

ryanhill1 (changes requested):
This fix seems to be working most of time, but after generating multiple plots with randomly chosen seeds, I also got these plots (among others) with nodes still cut off:

fbf70147-8064-4567-8dba-73869f15fe73

cb06c014-9d83-44e2-8f2e-139bbfe26a93

It seems like this solution only works for some of the possible seeds, not all.

Also, calling `graph.plot()` now generates the following warning:

```
.../lib/python3.11/site-packages/IPython/core/pylabtools.py:170: UserWarning: There are no gridspecs with layoutgrids. Possibly did not call parent GridSpec with the "figure" keyword
  fig.canvas.print_figure(bytes_io, **kw)
```

Is there anything we can do to address this? Suppressing the warning is an option, but ideally we’d find a proper fix if we move forward with this implementation.

SaashaJoshi:
I will look into this!

PR #993, opened 2025-06-08
Add margins for conv graph plot 

### LINKED ISSUE #851: [BUG] Fix nodes being cut-off around outside of conv graph plot

AishSweety:
## Summary of changes

The autoscaling that matplotlib uses will sometimes cause the plot to be clipped. This is more likely to occur when using the "scatter" plot function with large symbols. This change does two things:

1) Explicitly creates a matplotlib figure object and associated Axes instance and passes the Axes instance to mpl_draw
2) After mpl_draw returns, a margin of 15% is added to the plot.

The margin around the plot allows the contents to be rendered without clipping.

Note that currently, the margin is hard coded at 15%. To make this more flexible it could be added as an extra keyword argument for the plot_conversion_graph function.

Potentially closes #851

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ Types of changes:
 ### Added
 - Added `CudaQKernel.serialize` method that converts cudaq program to QIR string for `run_input` compatible format for `QbraidDevice.submit`. ([#972](https://github.com/qBraid/qBraid/pull/972))
 - Added support for batch jobs for devices from Azure provider. The `AzureQuantumDevice.submit` method now accepts single and batched `qbraid.programs.QPROGRAM` inputs. ([#953](https://github.com/qBraid/qBraid/issues/953))
-- Added the "margin" argument to plot_conversion_graph to prevent possible clipping. ([#993](https://github.com/qBraid/qBraid/pull/993))
+- Added the 'margin' argument to 'plot_conversion_graph' to prevent possible clipping. ([#993](https://github.com/qBraid/qBraid/pull/993))
 
 ### Improved / Modified
 - Updated `TimeStamps` schema to auto-compute `executionDuration` from `createdAt` and `endedAt` if not explicitly provided. ([#983](https://github.com/qBraid/qBraid/pull/983))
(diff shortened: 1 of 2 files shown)


ryanhill1:
Here are some examples that show how reducing the margins causes the nodes to appear more crowded in the center. While this does fix the original issue of nodes being cut off, it introduces a new issue with overly tight spacing, which could be seen as equally problematic.

Example 1 before:

d925a284-1356-4fd2-9a67-d4ef5fb7eea1

Example 1 after:

caa729dc-97c7-44b6-b1d2-a52aefad20c4

Example 2 before:

64e3d741-db7d-4d5b-aae5-fd93994caf30

Example 2 after:

2c38f05b-2d8e-4c7f-8a4d-9a30edc5dfd0

ryanhill1 (changes requested):
Ok, thanks for looking into that. Instead of using `kwargs.pop` can you explicitly add `margin` to the kwargs and docstring of the `plot_conversion_graph` function, and update `CHANGELOG.md`? After that, this should be good to go.

ryanhill1 (changes requested):
Please update `CHANGELOG.md`, see https://github.com/qBraid/qBraid/pull/993#issuecomment-2953366501

ryanhill1:
Please include the changelog update in this PR, not a new one. And link back to this PR, not the issue.

ryanhill1 (dismissed):

PR #930, opened 2025-03-11
Populate basis gates in IBM Quantum provider backends

cosenal:
Consider the following snippet instantiating an IBM Quantum provider backend.
```python3
provider = QiskitRuntimeProvider()
device = provider.get_device("ibm_sherbrooke")
print(device.profile.basis_gates)
```
Output before PR:
```
None
```
Output after PR:
```
{'x', 'ecr', 'id', 'rz', 'sx'}
```

PR #1003, opened 2025-06-18
Populate basis gates in AWS Braket devices

cosenal:
The equivalent of #930, but for AWS Braket devices now.

Property from doc: https://docs.aws.amazon.com/braket/latest/developerguide/braket-constructing-circuit.html#verbatim-compilation

Tested with
```python3
provider = BraketProvider()
device = provider.get_device("arn:aws:braket:eu-north-1::device/qpu/iqm/Garnet")
print(device.profile.basis_gates)
```
Output:
```
{'cz', 'prx'}
```

### DIFF
--- a/qbraid/runtime/aws/provider.py
+++ b/qbraid/runtime/aws/provider.py
@@ -133,7 +133,7 @@ class BraketProvider(QuantumProvider):
         paradigm: dict = capabilities.get("paradigm", {})
         action: dict = capabilities.get("action", {})
         num_qubits = paradigm.get("qubitCount")
-        basis_gates = paradigm.get("nativeGateSet", {})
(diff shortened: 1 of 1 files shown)


qbraid/runtime/aws/provider.py
```diff
@@ -133,6 +133,7 @@ def _build_runtime_profile(
         paradigm: dict = capabilities.get("paradigm", {})
         action: dict = capabilities.get("action", {})
         num_qubits = paradigm.get("qubitCount")
+        basis_gates = paradigm.get("nativeGateSet", {})
```
ryanhill1, line 136:
```suggestion
        basis_gates = paradigm.get("nativeGateSet", None)
```

PR #1004, opened 2025-06-18
fix: alter raise of warning instead of value error for probability sum.

vprusso:
As discussed on the thread of [this issue](https://github.com/unitaryfoundation/metriq-gym/issues/386), we decided to demote the probability sum not being equal to 1 from a `ValueError` to a `UserWarning`.

### DIFF
--- a/qbraid/runtime/postprocess.py
+++ b/qbraid/runtime/postprocess.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 from math import isclose
 from typing import Any, Union
+import warnings
 
 
 def normalize_batch_bit_lengths(measurements: list[dict[str, int]]) -> list[dict[str, int]]:
@@ -255,7 +256,7 @@ def distribute_counts(probs: dict[Any, float], shots: int) -> dict[Any, int]:
         {0: 9, 1: 1}
     """
     if not isclose(sum(probs.values()), 1.0, rel_tol=1e-7):
-        raise ValueError("Probabilities must sum to 1.")
+        warnings.warn("Probabilities must sum to 1.")
 
     if not all(0 <= prob <= 1 for prob in probs.values()):
         raise ValueError("Probabilities must be between 0 and 1.")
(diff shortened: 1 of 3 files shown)


qbraid/runtime/postprocess.py
```diff
@@ -255,7 +256,7 @@ def distribute_counts(probs: dict[Any, float], shots: int) -> dict[Any, int]:
         {0: 9, 1: 1}
     """
     if not isclose(sum(probs.values()), 1.0, rel_tol=1e-7):
-        raise ValueError("Probabilities must sum to 1.")
+        warnings.warn("Probabilities must sum to 1.")
```
ryanhill1, line 259:
```suggestion
        warnings.warn("Probabilities do not sum to 1.")
```

tests/runtime/test_result.py
```diff
@@ -816,7 +817,7 @@ def test_distribute_counts_probabilities_not_sum_to_1():
     """Test distribute_counts with probabilities that do not sum to 1."""
     probs = {0: 0.9, 1: 0.2}
     shots = 10
-    with pytest.raises(ValueError, match="Probabilities must sum to 1."):
+    with pytest.warns(UserWarning, match="Probabilities must sum to 1."):
```
ryanhill1, line 820:
```suggestion
    with pytest.warns(UserWarning, match="Probabilities do not sum to 1."):
```

ryanhill1 (changes requested):

tests/runtime/test_result.py
```diff
@@ -25,7 +25,6 @@
 from qbraid.runtime.postprocess import (
     distribute_counts,
     format_data,
-    normalize_batch_bit_lengths,
```
ryanhill1, line 28: Need to add this back

ryanhill1 (changes requested):

tests/runtime/test_result.py
vprusso: Ah, I don't quite know why that was removed. Must have vim-keyed that one out of existence accidentally. Added back in 73c871c635a3d82b44598a7d63e4d95657badf22

ryanhill1 (approved):

PR #1016, opened 2025-07-11
Add changes for workspace filtering in `QbraidProvider`

TheGupta2012:
## Summary of changes

* [`qbraid/runtime/native/provider.py`](diffhunk://#diff-6f361948043f3272388d09bc56766913777e1264542ac8051de61649872ed083R226-R228): Added logic in the `get_devices` method to filter devices by the `AWS` vendor when the workspace is set to `aws`.
* [`qbraid/runtime/native/provider.py`](diffhunk://#diff-6f361948043f3272388d09bc56766913777e1264542ac8051de61649872ed083R251-R255): Updated the `get_device` method to raise a `ResourceNotFoundError` if the requested device is not available in the current `AWS` workspace.

### DIFF
--- a/qbraid/runtime/native/provider.py
+++ b/qbraid/runtime/native/provider.py
@@ -223,6 +223,9 @@ class QbraidProvider(QuantumProvider):
                 and device["provider"] in {"Azure", "IonQ", "Quantinuum", "Rigetti", "Pasqal"}
             )
         ]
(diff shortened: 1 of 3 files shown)


TheGupta2012:
Yea makes sense, the CLI would need to be updated too. Thanks for the suggestion, will update core and cli as well!

Issue #1027, opened 2025-08-10
[BUG] Running circuits with measurement on IonQ through Braket


yitchen-tim:
### Environment - Output of qbraid.about()

[This block](https://github.com/qBraid/qBraid/blob/ecd07ce6900e4dfdf1fa9737028d636dde2febc0/qbraid/runtime/aws/device.py#L115) cause circuits with measurement fail. I think of two ways to solve it:
- Implement translation of measurement in `pytket-braket`. But Braket does not fully support classical register, so it is tricky to implement it. 
- Remove this block. I tried running few circuits on IonQ without this block and I have no issue so far. It's also unclear to me about the functionality of this code block. Can we remove this block? 

### Context - Running in qBraid Lab?

Yes

### What happened?

Running circuits with measurement on IonQ fails

ryanhill1:
@yitchen-tim Is this now resolved with https://github.com/qBraid/qBraid/pull/1028 ?

yitchen-tim:
Yes, resolved. I am closing this issue.

PR #1028, opened 2025-08-10
Support partial measurement on IonQ devices and Amazon Braket simulators

yitchen-tim:
## Summary of changes

This PR adds support for partial measurements on IonQ and Amazon Braket simulators, allowing users to measure only specific qubits in their quantum circuits while maintaining compatibility with existing workflows.

Circuit Processing (qbraid/programs/gate_model/braket.py)
- Added pad_measurements() method to detect existing measurements and pad circuits with measurements on all qubits
- Enhanced transform() method to automatically apply measurement padding for IonQ and Amazon Braket devices
- Added tracking of originally measured qubits via partial_measurement_qubits attribute

Device Submission (qbraid/runtime/aws/device.py)
- Enhanced submit() method to extract partial measurement information from circuits
- Added automatic tagging of quantum tasks with partial measurement qubit indices
- Implemented tag-based tracking for result processing

Job Management (qbraid/runtime/aws/job.py)
- Added _get_partial_measurement_qubits_from_tags() method to retrieve partial measurement information from AWS tags
- Enhanced result() method to pass partial measurement context to result builders
- Integrated with AWS Braket client for tag retrieval

Result Processing (qbraid/runtime/aws/result_builder.py)
- Enhanced BraketGateModelResultBuilder constructor to accept partial measurement qubit indices
- Updated measurements() and get_counts() methods to filter results for originally measured qubits
- Added utility functions marginal_measurement() and marginal_count() for result marginalization
- Enhanced BraketAhsResultBuilder for interface consistency

## Note
There is a TODO tag in the PR which requires a decision from this issue: https://github.com/qBraid/qBraid/issues/1027

## Example usage
```
from braket.circuits import Circuit

# Create circuit with partial measurements
circuit = Circuit().h(0).cnot(0, 1).cnot(1, 2).measure(0).measure(2)

# Submit to Amazon Braket simulator - automatic padding and tracking
provider = BraketProvider()
device = provider.get_device("arn:aws:braket:::device/quantum-simulator/amazon/sv1")
job = device.run(circuit, shots=1000)

# Results automatically filtered to show only qubits 0 and 2
result = job.result()
counts = result.data.measurement_counts  # Only contains 2-qubit bitstrings, even the 3 qubits are measured on the provider
```


qbraid/runtime/aws/device.py
```diff
@@ -112,7 +113,17 @@ def transform(
                 program, str(AnalogHamiltonianSimulation), experiment_type
             )
 
-        if provider == "IONQ":
+        ## TODO: This block causes error for circuits with measurement. Needs to restructure this block
+        ## add a tighter condition on when it is triggered.
```
ryanhill1, line 117: This might be a blocker. Can you expand on what you mean here in terms of the errors raised for circuits with measurement?

yitchen-tim:
After removing that block, some unit tests failed and I found that the block actually helps supporting a bigger set of gates. For example, CCX is not supported by the device, but the block converts these gates to supported gates. For backward compatibility, I think it's better leave it. I added a tighter constraint that bypass the block if there is a measure in the circuit (in which case the block will fail). I also added comment about why we need the block and its behavior.

ryanhill1:
@yitchen-tim Sounds good, thanks! I can take a look later on today.

ryanhill1 (approved):

ryanhill1:
Updates available in pre-release: https://pypi.org/project/qbraid/0.9.9.dev20250812233542/

PR #1059, opened 2025-10-02
Pad measurement to non-contiguous qubits for IonQ device

yitchen-tim:
## Summary of changes
This PR improves the measurement handling for circuits when running on IonQ devices and Braket simulators by ensuring all qubits in the range [0, N] are measured, where N is the highest qubit index used in the circuit. This is to satisfy the device's requirement that all qubit from [0, N] must be measured. (That is, even if qubit 0 and qubit 2 have gates, qubit 1 needs to be measured too).


ryanhill1 (approved):
LGTM

PR #1100, opened 2026-01-02
Handle region and endpoint from environ for AWS devices

yitchen-tim:
## Summary of changes

Amazon Braket SDK gets default region name and endpoint url from environment variables. This PR adds proper handle region and endpoint from environ for AWS devices. With this PR, users can now use environment variables to define region name (`"AWS_REGION"`) and endpoint url (`"BRAKET_ENDPOINT"`). This is useful when an application wraps over qBraid and does not have direct access to qBraid Provider class or AwsSession class. 

**Example use case**
```
import os

os.environ["AWS_REGION"] = "us-east-1"
os.environ["BRAKET_ENDPOINT"] = "https://braket.us-east-1.amazonaws.com"

# creating provider and device from specified region and endpoint
provider = BraketProvider()
device = provider.get_device("arn:aws:braket:::device/quantum-simulator/amazon/sv1")
```

### DIFF
--- a/qbraid/runtime/aws/job.py
+++ b/qbraid/runtime/aws/job.py
@@ -20,7 +20,6 @@ from __future__ import annotations
 
 from typing import TYPE_CHECKING, Optional
 
-import boto3
 from braket.aws import AwsQuantumTask
 from braket.tasks.analog_hamiltonian_simulation_quantum_task_result import (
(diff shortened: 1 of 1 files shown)


yitchen-tim:
I ran `tox -e linters` instead of `tox -e format-check`. So I missed an unused import. Just fixed it.

ryanhill1 (approved):

PR #1112, opened 2026-02-03
Add: legacy jobs retrieval , Fix: format with `black`

TheGupta2012:
## Summary of changes

Upstream Dependency : https://github.com/qBraid/qbraid-api/pull/368 should be merged before we deploy this 

**Legacy job management functionality:**

* Added `download_legacy_jobs` method to retrieve and extract a user's legacy jobs archive (`.old_jobs.json.zip`) from qBraid storage, with warnings about its temporary nature.
* Added `display_legacy_jobs` method to present legacy jobs in a table, automatically downloading the archive if not found locally, and displaying results with filtering options.


TheGupta2012:
Okay yeah, adding to core makes much more sense and was thinking that as well - will do those updates

Issue #211, opened 2023-05-16
Cirq transpiler will decompose gate itself


junliangtan1:
### Environment

- **Python version**:
- **Operating system**:

### What happened?

Cirq will decompose gate during transpile

### What is the expected outcome?

It shouldn't transpile if its framework support gate
or we can create a passes can create custom gate

junliangtan1:
@ryanhill1 to solve this issue and #59, I am thinking trying to add qasm custom operation support to qasm parser, so we can deprecate the decomposition you mention in https://github.com/qBraid/qBraid/issues/162#issuecomment-1478464877 and solve the decomposition problem and adding more circuit support, to implement this we can take reference from [qiskit ](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/qasm/qasmparser.py)and [pytket](https://github.com/CQCL/tket/blob/develop/pytket/pytket/qasm/qasm.py) qasm parsers.
what do you think about this?

ryanhill1:
@junliangtan1 Works for me. I agree that adding support for custom qasm gates / the ability to process gate definitions in the cirq parser would be a great improvement. Compared to the original cirq [`_parser.py`](https://github.com/quantumlib/Cirq/blob/master/cirq-core/cirq/contrib/qasm_import/_parser.py), our version has added support for `crx`, `cs`, `ccz`, `csdg`, `crz`, `cry`, `csx`, `c3sx`, `u`, `cu`, `cu1`, `cu3`, `p`, `cp`, `iswap`, and `rzz` gates. But there are still a good handful from the qiskit qasm [`qelib1.inc`](https://github.com/Qiskit/qiskit-terra/blob/main/qiskit/qasm/libs/qelib1.inc) that are not covered. Not all of the above mentioned gates are supported by cirq, so there's already a small collection of custom cirq gates that we've defined in [`custom_gates.py`](https://github.com/qBraid/qBraid/blob/main/qbraid/transpiler/custom_gates.py) that we import into the qasm parser for the conversion to avoid further decompositions where it isn't necessary. There is obviously many gates that are still not accounted for missing, so perhaps modeling after the more sophisticated qiskit and pytket qasm parsers could be a good direction. Alternatively, we could also keep our own copy of the `qelib1.inc` and `stdgates.inc` files and use them to perform a gate synthesis on the other end. But this may end up being more complicated. Happy to talk more about this.

PR #1116, opened 2026-02-07
Add: legacy jobs support for `QbraidProvider`

TheGupta2012:
## Summary of changes
Dependency : https://github.com/qBraid/qbraid-core/pull/211
Add the legacy jobs support to original provider, rather than using the `QuantumRuntimeClient`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,45 @@ Types of changes:
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


TheGupta2012:
@ryanhill1 the two failing tests are due to the fact that qbraid-core `QuantumClient` is not updated and we are using it in the test. Once you merge #211 and release core, it should be fixed

PR #992, opened 2025-06-04
Rigetti runtime

dhanaabhirajk:
## Summary of changes

This PR adds rigetti provider using [qcs-api-client](https://pypi.org/project/qcs-api-client/) and [pyquil](https://pyquil-docs.rigetti.com/). Potentially closes https://github.com/qBraid/qBraid/issues/981

Changes in 

- [x] Add Quantum Provider for rigetti
- [x] Add Quantum Device ( submit func is with list of programs)
- [x] Add Quantum Job ( yet to rebuild the result for qBraid result)
- [x] Add Unit tests
- [x] Lint checks

PR #1127, opened 2026-02-27
Add: Rigetti QCS runtime provider (RigettiProvider, RigettiDevice, RigettiJob)

### LINKED ISSUE #981: [FEATURE] Rigett QCS runtime integration
Create `qbraid.runtime.rigetti` module that implements provider, device, job, and result classes using the [qcs-api-client](https://pypi.org/project/qcs-api-client/).

Rigetti QCS Resources:
- https://github.com/rigetti/qcs-api-client-python
- https://qcs-api-client-python.readthedocs.io/en/stable/index.html

qBraid Runtime Resources:
- https://docs.qbraid.com/sdk/user-guide/runtime/new-provider

TheGupta2012:
Building off from #992 and fixes #981 

**New Feature** 

Implements the Rigetti QCS runtime provider for the qBraid SDK, adding `RigettiProvider`, `RigettiDevice`, and `RigettiJob` classes that follow the standard qBraid runtime interface. All QCS SDK interfaces are used directly (via `qcs_sdk`) to avoid requiring a local quilc/QVM installation.

> **WIP Note:** The quilc compilation step is not yet integrated. Programs submitted via `RigettiDevice.submit()` must already be expressed in native Quil gates (`RZ`, `RX`, `CZ`, `MEASURE`). Compilation support will be added in a follow-up.

## Changes

- **`RigettiProvider`**: new provider class using `qcs_sdk.client.QCSClient` for authentication; supports OAuth session via `RIGETTI_REFRESH_TOKEN`, `RIGETTI_CLIENT_ID`, and `RIGETTI_ISSUER` env vars, or a pre-constructed `QCSClient` passthrough; implements `get_devices()` and `get_device()` backed by `list_quantum_processors` and `get_instruction_set_architecture`
- **`RigettiDevice`**: new device class wrapping a single Rigetti QPU or simulator; implements `status()` (using `get_instruction_set_architecture` to probe liveness), `live_qubits()`, and `submit()` (translate + submit via `qcs_sdk.qpu.api`)
- **`RigettiJob`**: new job class backed by `qcs_sdk.qpu.api`; implements `status()`, `cancel()`, `get_result()`, and `result()` with readout buffer parsing into `GateModelResultData`
- **`pyproject.toml`**: added `rigetti` optional dependency group pinned to `pyquil>=4.16.2`
- **`requirements-dev.txt`**: bumped `pyquil` to `>=4.16`, removed the old pinned `qcs-sdk-python` range, and added `qcs-api-client>=0.26.5`
- **`tox.ini`**: pass-through `RIGETTI_REFRESH_TOKEN`, `RIGETTI_CLIENT_ID`, and `RIGETTI_ISSUER` env vars in CI
- **`tests/`**: unit tests for provider init (env var validation), device status/live-qubits, job lifecycle (submit, cancel, result), and result parsing using mocks
- **`docs/`**: added `rigetti` to the runtime API reference autodoc index

## References

- Rigetti QCS SDK (Python): https://github.com/rigetti/qcs-sdk-rust/tree/main/crates/python
- qBraid runtime interface: `qbraid.runtime.QuantumProvider`, `qbraid.runtime.QuantumDevice`, `qbraid.runtime.QuantumJob`

## Summary by CodeRabbit

* **New Features**
  * Rigetti provider integrated: discover devices, submit jobs, monitor status, and retrieve results from Rigetti QCS.

* **Authentication**
  * Added Rigetti QCS environment variables (refresh token, client id, issuer) to examples and test environments.

* **Documentation**
  * API docs updated to surface the Rigetti runtime.

* **Chores**
  * Optional Rigetti runtime dependency and provider/job entry points added.

* **Tests**
  * Comprehensive unit tests for provider, device, and job behaviors.


ryanhill1 (changes requested):

ryanhill1 (commented):
Make sure to update `docs/api/qbraid.runtime.rst` by adding the `rigetti` sub-module

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,137 @@
...
+            warnings.warn(
+                UserWarning(
+                    "Failed to cancel the QPU job. "
+                    "Cancellation is not guaranteed, as it "
+                    "is based on job state at the time of cancellation."
+                )
+            )
+            self._status = JobStatus.RUNNING
+        else:
+            self._status = JobStatus.CANCELLED
```
ryanhill1: don't hard-code status value

ryanhill1 (changes requested):

ryanhill1:
This being built on top of https://github.com/qBraid/qBraid/pull/992 we may want to rebase on top of that PR eventually in order to give that contributor due credit.

qbraid/runtime/rigetti/device.py
```diff
@@ -0,0 +1,193 @@
...
+            quantum_processor_ids = set(list_quantum_processors(client=self._qcs_client))
+            if self.id not in quantum_processor_ids:
+                return DeviceStatus.OFFLINE
+        except ListQuantumProcessorsError as e:
+            raise RigettiDeviceError(
+                "Failed to retrieve quantum processor list from Rigetti QCS."
+            ) from e
+        return DeviceStatus.ONLINE
+
+    def transform(self, run_input):
```
ryanhill1: Add a type annotation. Also, this should only accept one input type with a congruent output type matching what is expected in the runtime flow.

qbraid/runtime/rigetti/device.py
```diff
@@ -0,0 +1,193 @@
...
+    def _submit(self, run_input: str, shots: Optional[int] = None) -> RigettiJob:
+        """
+        Submit a Quil program to the Rigetti QPU.
+
+        Args:
+            run_input: A serialized Quil program string (produced by prepare()).
+            shots: Number of shots for the job.
+        """
+        num_shots = shots
+        if num_shots is None or num_shots <= 0:
```
ryanhill1: if shots is required it shouldn't be optional in function signature

qbraid/runtime/rigetti/device.py
```diff
@@ -0,0 +1,193 @@
...
+            device=self,
+            num_shots=num_shots,
+            ro_sources=translation_result.ro_sources,
+        )
+
+    # pylint: disable-next=arguments-differ
+    def submit(
+        self,
+        run_input: Union[str, list[str]],
+        shots: Optional[int] = None,
```
ryanhill1: it shots actually optional? or is it required?

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+    """Class for errors raised while processing an Rigetti job."""
+
+
+class RigettiJob(QuantumJob):
+    """Rigetti job class."""
+
+    def __init__(
+        self,
+        job_id: str | int,
+        device: RigettiDevice,
```
ryanhill1: Should be optional `QCSClient` argument instead to match the convention of the other Job classes. Device can also be passed but should be optional

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+class RigettiJobError(QbraidRuntimeError):
+    """Class for errors raised while processing an Rigetti job."""
+
+
+class RigettiJob(QuantumJob):
+    """Rigetti job class."""
+
+    def __init__(
+        self,
+        job_id: str | int,
```
ryanhill1: Are the job ids str or int? I doubt it is both

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+        device: RigettiDevice,
+        num_shots: int = 1,
```
ryanhill1: what's the reason for this placeholder value? Could create confusion

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+        ro_sources: Optional[dict] = None,
```
ryanhill1: annotate the type of `dict` e.g. `dict[str, str]` ? There's no docstring either so not sure what this is

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+        **kwargs: Any,
+    ):
+        super().__init__(job_id=job_id, device=device, **kwargs)
+        self._num_shots = num_shots
+        self._ro_sources = ro_sources or {}
+        self._status = JobStatus.INITIALIZING
+        self._cached_results = None
+
+    @property
+    def _client(self):
```
ryanhill1: this can be a public method

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+        return self._device._qcs_client
+
+    @property
+    def execution_duration_microseconds(self) -> int | None:
```
ryanhill1: would it make more sense for this to just be passed within result details rather than exposed via job class? Bc it is only available after result has been retrieved, not as part of job data

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+                "is based on job state at the time of cancellation."
+            ) from exc
+        self._status = JobStatus.CANCELLED
+        logger.info(
+            "Successfully cancelled Rigetti job %s on device %s",
+            self.id,
+            self._device.id,
+        )
+
+    def _build_register_map(self, execution_results):
```
ryanhill1: Add a type annotation. What is the type of execution_results and what is the return type?

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+            key: ReadoutValues(value.data) for key, value in execution_results.buffers.items()
+        }
+        qpu_result_data = QPUResultData(
+            mappings=dict(self._ro_sources),
+            readout_values=readout_values,
+            memory_values=execution_results.memory,
+        )
+        return qpu_result_data.to_register_map()
+
+    def _parse_results(self, execution_results) -> GateModelResultData:
```
ryanhill1: type annotation for execution results

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+        all_bits = np.hstack(register_arrays)
+
+        measurements = []
+        for row in all_bits:
+            measurements.append("".join(str(b) for b in row))
+
+        counts = {m: measurements.count(m) for m in set(measurements)}
+        return GateModelResultData(measurement_counts=counts)
+
+    def result(self, timeout: Optional[int] = None) -> Result:
```
ryanhill1: can annotate as `Result[GateModelResultData]` for better type hints I think

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+        """Return the result of the Rigetti job.
+
+        Raises:
+            RigettiJobError: If the job result cannot be retrieved or parsed.
```
ryanhill1, line 251: I'm guessing this is raised within `_parse_results`? I'm not sure what the standard is but it might be more idomatic to catch and re-raise to match the error handling within the function to the annotation. Otherwise `_parse_results` could change and cause this docstring to be misaligned with the expected behavior

qbraid/runtime/rigetti/job.py
```diff
@@ -0,0 +1,239 @@
...
+            )
+        result_data = self._parse_results(self._cached_results)
+
+        self._status = JobStatus.COMPLETED
+        return Result[GateModelResultData](
+            device_id=self._device.id,
+            job_id=self.id,
+            success=True,
+            data=result_data,
+        )
```
ryanhill1, line 281: additional kwargs passed here will get stored in `result.details` so you could pass the execution duration here and anything else applicable

qbraid/runtime/rigetti/provider.py
```diff
@@ -0,0 +1,295 @@
...
+
+        # --- Rebuild QCSClient ---
+        self._qcs_client = build_qcs_client(
+            refresh_token=refresh_token,
+            client_id=client_id,
+            issuer=issuer,
+            grpc_api_url=grpc_api_url,
+            quilc_url=quilc_url,
+            qvm_url=qvm_url,
+        )
```
ryanhill1: This seems redundant, don't we basically already do this within the `__init__` method? Is there a reason we want to repass each of these variables and recreate the client, separate from what was either passed to the provider or created on init?

CHANGELOG.md
ryanhill1, line 1: please add usage example to changelog

ryanhill1 (changes requested):
If you try to call `device.run(...)` without one of the necessary containers running, the call will hang indefinitely. This could be an issue within the runtime api if a container exits unexpectedly. We will want to raise an exception quickly if the quilc container is not running to avoid this.

Please replace all instances of `Optional[x]` with `x | None`, as this is the convention we are slowly trying to move towards throughout the SDK.

I also believe the `RigettiDevice` class needs a `__str__` method to match the convention of the others e.g. `return f"{self.__class__.__name__}('{self.id}')"`.

Also, should we add `execution_options` as a kwarg to the `submit` method instead of as a kwarg to the device class? What different options are available there?

This branch, in it's current state, is now preleased under `qbraid==0.11.2.dev20260416024038` so that you can test with runtime api as well.

qbraid/runtime/rigetti/device.py
```diff
@@ -0,0 +1,128 @@
...
+
+        # TODO: figure out how to add the quilc compilation step here
+        #       Either we need to host that compiler or there must be
+        #       some way for us to use it in the local system
+        translation_result = translate(
+            native_quil=quil_program,
+            num_shots=num_shots,
+            quantum_processor_id=quantum_processor_id,
+            client=self._qcs_client,
+        )
```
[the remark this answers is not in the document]
ryanhill1: Shots can be included in a pyquil program though right? I'm just thinking about being able to control `device.set_options(transform=False)` to disable circuit transformations, whereas keeping this in the `.run` method doesn't allow the user any control over this "pass". Having a method signature that differs from the parent class isn't a huge deal.

qbraid/runtime/rigetti/provider.py
TheGupta2012: This is required because if a user were to pass in new URLs for the compilation endpoints, the client object would change

ryanhill1 (commented):
Resolve merge conflicts and then this should be g2g

qbraid/runtime/rigetti/device.py
```diff
@@ -0,0 +1,375 @@
...
+            translation_options=translation_options,
+        )
+
+    def run(
+        self,
+        run_input: Union[pyquil.Program, list[pyquil.Program]],
+        *args,
+        runtime_options: dict[str, Any] | None = None,
+        **kwargs,
+    ) -> Union[RigettiJob, list[RigettiJob]]:
```
ryanhill1: I must have missed this in my first review. Why are we overriding the parent class `run` method?

TheGupta2012: This was to introduce the `runtime_options` dict in the `run` method signature. Moreover, the original `run` method does not allow us to [pass in the rigetti `translation_options` inside the `submit` method](https://github.com/qBraid/qBraid/pull/1127/changes/BASE..8703e359456874b6064d3587deeeda5bab29d4d0#diff-a98fdb0cc83e3b027639c3fc89372e46cfb6a9b3f97deb1704944f08b89a5a6fR360). Were you thinking of bundling the `runtime_options` dict as part of the `kwargs`?

ryanhill1: yeah all of that can be passed as part of kwargs from the parent run method to the rigetti submit method

ryanhill1 (approved):

PR #1140, opened 2026-03-18
Add BatchJobSession context manager for grouped execution

TheGupta2012:
**New group job submission feature:**

* Added `GroupJobSession` context manager and `GroupResult` container to `qbraid.runtime`, enabling logical grouping of quantum jobs. Jobs submitted via `QbraidDevice.run()` within an active group session are automatically tagged and registered; supports both context-manager and manual open/close usage, optional TTL, and completion callbacks. Also exports `get_active_group` for retrieving the current group QRN. [[1]](diffhunk://#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR19-R88) [[2]](diffhunk://#diff-3eeb31f36e47d8204ecedeb7451f2c4efd214607f612aaa1a7e1396559fa0105R91-R99) [[3]](diffhunk://#diff-3eeb31f36e47d8204ecedeb7451f2c4efd214607f612aaa1a7e1396559fa0105R116-R118)

**Integration with native device/job submission:**

* Updated `QbraidDevice.submit()` to check for an active group session, automatically tag jobs with the group QRN, and register jobs with the session, ensuring SDK and backend state remain consistent. [[1]](diffhunk://#diff-c12c9b3a861ce10e8c8e44d8a880e620c4fc5d5f3e5d62e40f6b09b31e51928cR21-R31) [[2]](diffhunk://#diff-c12c9b3a861ce10e8c8e44d8a880e620c4fc5d5f3e5d62e40f6b09b31e51928cL120-R148) [[3]](diffhunk://#diff-c12c9b3a861ce10e8c8e44d8a880e620c4fc5d5f3e5d62e40f6b09b31e51928cR159-R165)

**Test infrastructure updates:**

[...]

PR #1148, opened 2026-03-26
Add OriginQ QCloud provider

vprusso:
Add native support for OriginQ Wukong quantum devices via the pyqpanda3 SDK.

- OriginDevice: QProg conversion from QASM/Qiskit, simulator/hardware submission paths
- OriginJob: status mapping, result retrieval with counts fallback
- Thread-safe QCloudService caching per API key
- 27 unit tests covering provider, device, job, and status mapping

Closes qBraid/qBraid#1146

FYI: Integration in metriq-gym (currently using this branch to dispatch/poll job on 180-qubit superconducting qubit OriginQ device):
https://github.com/unitaryfoundation/metriq-gym/pull/735

## Summary by CodeRabbit

* **New Features**
  * Added OriginQ provider: discover devices, submit single or batch circuits, monitor jobs, and retrieve results.

* **Documentation**
  * OriginQ runtime submodule added to API docs and changelog.

* **Tests**
  * New tests for API key resolution, provider/device discovery, job status mapping, submission, and result handling.

* **Chores**
  * Added optional OriginQ dependency and provider registration for runtime discovery.


ryanhill1:
@vprusso Thanks for this PR! I will try to review within the next few days

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+    def backend_name(self) -> str:
+        """Return the OriginQ backend name."""
+        return self._backend_name
+
+    def status(self) -> DeviceStatus:
+        """Return the current status of the OriginQ device."""
+        # OriginQ does not expose health endpoints; assume online when instantiated.
+        return DeviceStatus.ONLINE
+
+    def _to_qprog(self, run_input):
```
ryanhill1: We already have a [`qasm2_to_pyqpanda3`](https://github.com/qBraid/qBraid/blob/6be2631d0a5cbe768fc1978989045743bc38330a/qbraid/transpiler/conversions/qasm2/qasm2_extras.py#L78) function in the transpiler, so this method is not needed.

The qiskit to qasm2 conversion will also be handled automatically by the transpiler once the target program type is set to pyqpanda3.

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+        backend_name: str,
+    ) -> TargetProfile:
+        """Build a TargetProfile from an OriginQ backend."""
+        simulator = backend_name in SIMULATOR_BACKENDS
+        return TargetProfile(
+            device_id=device_id,
+            simulator=simulator,
+            experiment_type=ExperimentType.GATE_MODEL,
+            num_qubits=_infer_num_qubits(backend, backend_name, simulator=simulator),
+            program_spec=ProgramSpec(str, alias="qasm2"),
```
ryanhill1: If the goal is to submit the program as a `pyqanda3.core.QProg` program type, then that is what should be set here in order for the transpiler to handle it automatically.

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+            return convert_qasm_string_to_qprog(run_input)
+
+        raise TypeError(
+            f"Unsupported input type {type(run_input)}; "
+            "expected OpenQASM 2.0 string or qiskit.QuantumCircuit"
+        )
+
+    def transform(self, run_input):
+        """Transform a single input to OriginQ's QProg format."""
+        return self._to_qprog(run_input)
```
ryanhill1: The transform method is intended to be used for circuit transformations (e.g. decompositions) within a given program type, not translations between program types. Since the transpiler will handle conversion, this transform method can be removed. By default it is a no-op

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+    def run(self, run_input, *args, **kwargs):
+        """Run a quantum program or list of programs on this device."""
+        is_single = not isinstance(run_input, Sequence) or isinstance(run_input, str)
+        inputs = [run_input] if is_single else run_input
+        jobs = [self.submit(self._to_qprog(item), *args, **kwargs) for item in inputs]
+        return jobs[0] if is_single else jobs
+
+    def submit(self, run_input, *, shots: int, **kwargs) -> OriginJob:
+        """Submit a single quantum program to the OriginQ device."""
+        qprog = self._to_qprog(run_input)
```
ryanhill1: After configuring the `ProgramSpec` correct, you can assume that by the time the `submit` method is reached, the `run_input` will be of type `QProg | list[QProg]`.

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+    def transform(self, run_input):
+        """Transform a single input to OriginQ's QProg format."""
+        return self._to_qprog(run_input)
+
+    def run(self, run_input, *args, **kwargs):
+        """Run a quantum program or list of programs on this device."""
+        is_single = not isinstance(run_input, Sequence) or isinstance(run_input, str)
+        inputs = [run_input] if is_single else run_input
+        jobs = [self.submit(self._to_qprog(item), *args, **kwargs) for item in inputs]
+        return jobs[0] if is_single else jobs
```
ryanhill1: This is handled by the parent `QuantumDevice` class. Overriding the parent `run` method would disable the `transpile` $\rightarrow$ `transform` $\rightarrow$ `validate` steps. This method can be removed and left to the parent class.

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+            backend_job = self._backend.run(qprog, nshots)
+        else:
+            # pylint: disable-next=import-outside-toplevel
+            from pyqpanda3 import qcloud as qcloud_module
+
+            options = qcloud_module.QCloudOptions()
+            backend_job = self._backend.run(qprog, nshots, options)
+
+        job_id = backend_job.job_id()
+        return OriginJob(
```
ryanhill1: should also return `list[OriginJob]` if `run_input` is of type `list[QProg]`

qbraid/runtime/origin/job.py
```diff
@@ -0,0 +1,138 @@
...
+    "FAILED": JobStatus.FAILED,
+    "ERROR": JobStatus.FAILED,
+}
+
+
+class OriginJobError(QbraidRuntimeError):
+    """Class for errors raised while processing an OriginQ job."""
+
+
+def _normalize_status(origin_status: Any) -> str:
```
ryanhill1, line 45: what is the type of `origin_status`? Should declare it explicitly rather than `Any`

qbraid/runtime/origin/job.py
```diff
@@ -0,0 +1,138 @@
...
+
+        counts = result_obj.get_counts()
+        if not counts:
+            try:
+                counts_list = result_obj.get_counts_list()
+            except Exception:
+                counts_list = []
+            if counts_list:
+                counts = counts_list[0]
+        counts = counts or {}
```
ryanhill1, line 130: so is it possible for a job to be successful but not return any measurement counts?

qbraid/runtime/origin/job.py
```diff
@@ -0,0 +1,138 @@
...
+
+        normalized = {str(key): int(value) for key, value in counts.items()}
+        device_id = self._device.profile.device_id if self._device else "origin"
```
ryanhill1, line 133: there's no job metadata that tells you which device the job was run on?

qbraid/runtime/origin/provider.py
```diff
@@ -0,0 +1,110 @@
...
+    """Return a cached QCloudService instance for the given API key."""
+    # pylint: disable-next=import-outside-toplevel
+    from pyqpanda3 import qcloud as qcloud_module
+
+    with _SERVICE_LOCK:
+        service = _SERVICE_CACHE.get(api_key)
+        if service is None:
+            service = qcloud_module.QCloudService(api_key=api_key)
+            _SERVICE_CACHE[api_key] = service
+    return service
```
ryanhill1: Can you help me understand the need for this vs just returning `qcloud_module.QCloudService(api_key=api_key)`?

qbraid/runtime/origin/job.py
```diff
@@ -0,0 +1,138 @@
...
+class OriginJob(QuantumJob):
+    """OriginQ QCloud job class."""
+
+    def __init__(
+        self,
+        job_id: str,
+        *,
+        device: Optional[Any] = None,
+        backend_job: Optional[Any] = None,
+        api_key: Optional[str] = None,
```
ryanhill1: to align with the convention of the other provider job classes, we should (optionally) be passing `QCloudService` here instead of api-key

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+
+class OriginDevice(QuantumDevice):
+    """OriginQ QCloud device interface."""
+
+    def __init__(
+        self,
+        profile: TargetProfile,
+        backend: QCloudBackend,
+        backend_name: str,
+        api_key: Optional[str] = None,
```
ryanhill1: optional param should be `QCloudService` instead of api_key

qbraid/runtime/origin/device.py
```diff
@@ -0,0 +1,167 @@
...
+    ):
+        super().__init__(profile=profile)
+        self._backend = backend
+        self._backend_name = backend_name
+        self._api_key = api_key
+
+    @staticmethod
+    def build_profile(
```
ryanhill1: we typically put this in the provider module

pyproject.toml
ryanhill1, line 1: let's also add an `origin` dependency extra with pyqpanda3

ryanhill1 (changes requested):
Good start! Gave a few notes on how we can more closely align this PR to leverage the existing infrastructure and align with the conventions used by the other provider classes.

Also, before merging will want to update the changelog and add `origin` line to `docs/api/qbraid.runtime.rst`. 

Also update [`qbraid/runtime/loader.py`](https://github.com/qBraid/qBraid/blob/6be2631d0a5cbe768fc1978989045743bc38330a/qbraid/runtime/loader.py) with origin provider

qbraid/runtime/origin/device.py
vprusso: Done — removed `_to_qprog` entirely. The transpiler's `qasm2_to_pyqpanda3` is now used via the `ProgramSpec(QProg, alias="pyqpanda3", serialize=qasm2_to_pyqpanda3)` in `_build_profile`.

qbraid/runtime/origin/device.py
vprusso: Done — `ProgramSpec` now uses `QProg` as the program type with `qasm2_to_pyqpanda3` as the serialize callback, so the transpiler handles conversion automatically.

qbraid/runtime/origin/device.py
vprusso: Done — removed `transform` override. The base class no-op is sufficient since the transpiler handles the QASM→QProg conversion.

qbraid/runtime/origin/device.py
vprusso: Done — `submit` now assumes `run_input` is `QProg | list[QProg]` and no longer calls `_to_qprog`.

qbraid/runtime/origin/device.py
vprusso: Done — removed the `run` override entirely. The parent `QuantumDevice.run` now handles the transpile → transform → validate → submit pipeline.

qbraid/runtime/origin/device.py
vprusso: Done — `submit` now handles both single `QProg` and `list[QProg]` inputs, returning `OriginJob` or `list[OriginJob]` accordingly.

qbraid/runtime/origin/job.py
vprusso: Good point — the pyqpanda3 QCloud SDK returns status as a string (e.g. "FINISHED", "COMPUTING"). The \_normalize\_status function handles both string and enum-like objects defensively since the SDK's return type isn't formally documented. Will investigate further whether there's a concrete enum type we can reference instead of `Any`.

qbraid/runtime/origin/job.py
vprusso: In practice, a successful job should always return counts. The `counts or {}` fallback is a defensive guard against edge cases where `get_counts()` might return `None` rather than an empty dict. Happy to change this to raise an error instead if that's preferred — e.g. `raise OriginJobError` when both `get_counts()` and `get_counts_list()` return empty for a completed job.

qbraid/runtime/origin/job.py
vprusso: Unfortunately, the QCloud SDK's job result object doesn't expose a device/backend identifier directly. The fallback to the device stored on the job object (or "origin" as default) is the best we have currently. If you know of a metadata field on the QCloud side I may have missed, happy to use it.

qbraid/runtime/origin/provider.py
vprusso: Removed the service cache. `_get_service` now simply returns a new `QCloudService(api_key=api_key)` each time. The provider lazily creates and stores its own service instance via the `service` property, so caching at the module level was redundant.

qbraid/runtime/origin/job.py
vprusso: Done — `OriginJob` now accepts an optional `service` (`QCloudService`) parameter instead of `api_key`, aligning with the IonQ convention of passing the session/service object.

qbraid/runtime/origin/device.py
vprusso: Done — `OriginDevice.__init__` now accepts `service: Optional[QCloudService]` instead of `api_key`.

qbraid/runtime/origin/device.py
vprusso: Done — moved `build_profile` to `OriginProvider._build_profile()` as a private method on the provider, consistent with IonQ's pattern.

pyproject.toml
vprusso: Done — added `origin = ["pyqpanda3"]` to `[project.optional-dependencies]` in pyproject.toml. Also added `origin` overloads to `loader.py`, added `origin` to `docs/api/qbraid.runtime.rst`, and updated the changelog.

ryanhill1:
@vprusso Thanks for the updates, I will try to re-review this week

vprusso:
Hey @ryanhill1 as I'm not super familiar with the coderabbit workflow, for these flagged issues, would you like me to take care of them, or are these taken care of by coderabbit? Just asking to make sure that I'm not blocking you here.

ryanhill1:
Yeah no problem, I can take the PR from here.

I still have to play with the coderabbit config a little. I think it's helpful for the initial pass but the automatic reviews on follow up commits add too much clutter.

vprusso:
> Yeah no problem, I can take the PR from here.
> 
> I still have to play with the coderabbit config a little. I think it's helpful for the initial pass but the automatic reviews on follow up commits add too much clutter.

Excellent, thanks @ryanhill1 . Feel free to let me know if I can help. Once this goes in, if it's okay with you, I may opt to do a similar type of upstream for Quantinuum as well. I'll wait until this goes in first before bombarding you with that though :)

ryanhill1:
Sounds good! I should be able to get this merged by EOD, and can pre-release from there.

Next official release will go out within the next two weeks after we merge in https://github.com/qBraid/qBraid/pull/1127 and https://github.com/qBraid/qBraid/pull/1140. Happy to include https://github.com/qBraid/qBraid/pull/1149 as well if we can get it turned around in time.

cosenal:
@ryanhill1 I will definitely address your comments on #1149 next week. Sorry I haven't had the time this week.

ryanhill1:
@cosenal No problem at all, been pretty busy around here too. No rush

ryanhill1:
@vprusso I made a few updates. Before we merge, can you confirm that everything still works for your use case as intended?

The most significant update is that the `submit` method now handles `list[QProg]` input differently depending on the backend type:

- **Simulator backends**: Returns `list[OriginJob]` — each circuit is submitted as an individual job since the pyqpanda3 SDK only supports native batch submission on QPU hardware.
- **QPU backends**: Returns a single `OriginJob` containing all programs, using the SDK's native batch API.

vprusso:
Thanks for the updates @ryanhill1! I tested this against our metriq-gym use case and found one regression:

**Polling jobs by ID no longer works** because `OriginJob._get_job` creates a bare `QCloudJob(job_id)` without initializing a `QCloudService` first. The pyqpanda3 SDK requires the service (which configures the HTTP URL) to be initialized before `QCloudJob.status()` or `.result()` calls will work.

Reproducer:
```python
from pyqpanda3 import qcloud

# Simulates what happens when qBraid's load_job() is called for an Origin job
# without a service (e.g., when polling a previously dispatched job by ID)
job = qcloud.QCloudJob("8BB5A3135B9A5D1A0F9039C8ED920014")
job.status()
# RuntimeError: curl_easy_perform() failed: URL using bad/illegal format or missing URL
```

With service initialized first, it works fine:
```python
service = qcloud.QCloudService(api_key=os.environ["ORIGIN_API_KEY"])
job = qcloud.QCloudJob("8BB5A3135B9A5D1A0F9039C8ED920014")
job.status()  # JobStatus.COMPUTING
```

This is the same issue addressed in [commit 5226e549](https://github.com/qBraid/qBraid/pull/1148/commits/5226e549c787085842f65ffb1edb613cc2cfb125) ("Fix: resolve API key from env when polling OriginQ jobs"), which got refactored away in the recent push.

**Suggested fix**: In `OriginJob._get_job` (or `OriginJob.__init__`), when no `job` and no `service` is passed, initialize a service from the env var before creating the `QCloudJob`. Something like:

```python
if service is None and job is None:
    from qbraid.runtime.origin.provider import _resolve_api_key, _get_service
    _get_service(_resolve_api_key())  # initializes the SDK's global state
```

End-to-end test results on our side:

| Operation | Result |
|-----------|--------|
| `provider.get_devices()` | ✅ works |
| `device.run(circuits, shots=N)` (dispatch) | ✅ works |
| `mgym job poll <job_id>` (polling by ID) | ❌ fails with URL error |

The batch handling changes look good for our use case — metriq-gym benchmarks that submit `list[QProg]` (like BSEQ, Mirror Circuits) will now get `list[OriginJob]` on simulators and a single `OriginJob` on QPUs, which is fine.

Happy to push a fix on this branch if you want, or leave it for you to handle.

ryanhill1:
@vprusso Thanks for pointing that out! Just pushed a fix. Let me know if that resolves it.

vprusso:
Fix confirmed — polling by ID now works correctly. I re-ran the metriq-gym dispatch/poll flow on the `full_amplitude` simulator with a BSEQ benchmark and polling returned proper `RUNNING` statuses across all sub-jobs instead of the previous URL error.

One tangential observation while testing (not a regression from this PR, just flagging): when a pyqpanda3 job genuinely fails on the cluster (e.g. `"the task was failed on cluster: 202"`), `QCloudJob.status()` raises a `RuntimeError` rather than returning a terminal `FAILED` status. Our current `status()` wraps that in `OriginJobError`, which means failed jobs can't be polled to completion — they just raise forever. Might be worth mapping that to `JobStatus.FAILED` in a follow-up, but not blocking for this PR.

LGTM from my end — feel free to merge.

ryanhill1:
Gotcha, I think that job failure bug is probably worth fixing. Do you want to push a fix for that since you have a failed job ID you can test with?

vprusso:
Done — pushed in [73cdfd15](https://github.com/qBraid/qBraid/commit/73cdfd15).

**Change summary**:
- `OriginJob.status()` now catches `RuntimeError` from the pyqpanda3 SDK, detects the "failed on cluster" substring, and returns `JobStatus.FAILED` (cached in `_cache_metadata`). Other `RuntimeError`s still propagate as `OriginJobError`.
- `OriginJob.result()` now checks status before calling `_job.result()` and raises a clean `OriginJobError("... failed on the cluster.")` for terminally-failed jobs instead of letting the pyqpanda3 `RuntimeError` bubble up.
- Added two unit tests covering both paths.

**Verified against the real failed job ID** (`8BB5A3135B9A5D1A0F9039C8ED920014`):
```
Test 1 - status(): JobStatus.FAILED  ✓
Test 2 - result() raised: OriginQ job 8BB5A3135B9A5D1A0F9039C8ED920014 failed on the cluster.  ✓
```

ryanhill1 (approved):

ryanhill1:
Thank you @vprusso for all of your work on this PR!

vprusso:
> Thank you @vprusso for all of your work on this PR!

My pleasure, @ryanhill1 ! Thank you for the comments and input on your side! If I haven't completely fatigued you here, I should be able to get a similar working MR up sometime this week for Quantinuum. Thanks, @ryanhill1 !

ryanhill1:
Sounds great. I'll be in DC for the DARPA HARQ kick-off meeting for the first part of this week, but can probably give a review Thursday, Friday, or over the weekend depending on what comes up.

These updates have been pre-released under `qbraid==0.11.2.dev20260420142908`: https://pypi.org/project/qbraid/0.11.2.dev20260420142908/ so you can feel free to test, and happy to merge in any further updates if needed.

PR #1160, opened 2026-04-14
Fix classical bit collisions in Braket pad_measurements

ryanhill1:
# Summary of Changes

## Problem

qBraid submits OpenQASM 3 programs to Braket via `Circuit.from_ir → pad_measurements → Circuit.to_ir`. Two flaws in that round trip can silently drop measurements:

**1. `Circuit.from_ir` reads classical bit indices literally from the source.** When a user's program uses multiple classical registers, Braket stores the literal bit index in each `Measure._target_index`. Two registers sharing index `0` produce two measures with `_target_index == 0`, and on re-serialization they overwrite the same bit.

**2. `Circuit.measure(q)` assigns `_target_index` from a running counter**, not from `q`. So padded measures occupy slots `[k, k+m-1]` — and if an existing measure already sits in that range, padding collides with it.

### Concrete example

Input from a real task (trimmed):

```qasm
OPENQASM 3.0;
qubit[16] q;
bit[2] sel;
bit[2] reg_A;
bit[2] reg_B;
bit[4] reg_P;
// ... gates ...
sel[0]   = measure q[0];
sel[1]   = measure q[1];
reg_A[0] = measure q[2];
reg_A[1] = measure q[3];
reg_B[0] = measure q[4];
reg_B[1] = measure q[5];
reg_P[0] = measure q[6];
reg_P[1] = measure q[7];
reg_P[2] = measure q[8];
reg_P[3] = measure q[9];
```

After `Circuit.from_ir`, the ten measures land with `_target_index = [0, 1, 0, 1, 0, 1, 0, 1, 2, 3]` — four collisions on slot `0`, four on slot `1`. `pad_measurements` then adds measurements for `q[10..15]` via `Circuit.measure()`, which assigns indices starting at `len(_measure_targets) == 10`, i.e. `[10, 11, 12, 13, 14, 15]`.

The QASM that Braket actually submits to IonQ Forte-1:

```qasm
bit[16] b;
...
b[0] = measure q[0];
b[1] = measure q[1];
b[0] = measure q[2];   // clobbers b[0]
b[1] = measure q[3];   // clobbers b[1]
b[0] = measure q[4];   // clobbers b[0]
b[1] = measure q[5];   // clobbers b[1]
b[0] = measure q[6];   // clobbers b[0] — survives
b[1] = measure q[7];   // clobbers b[1] — survives
b[2] = measure q[8];
b[3] = measure q[9];
b[10] = measure q[10];
b[11] = measure q[11];
b[12] = measure q[12];
b[13] = measure q[13];
b[14] = measure q[14];
b[15] = measure q[15];
```

Only the **last** write to each bit survives. The task runs and returns `measured_qubits = [6, 7, 8, 9, 10, 11, 12, 13, 14, 15]` — the original measurements of `q[0..5]` are silently lost, and the 10-bit result has nothing to do with what the user wrote.

On top of that, loading the corrupted task crashes:

```
File "qbraid/runtime/aws/job.py", line 190, in <listcomp>
    all_measurement_qubits.index(q) for q in partial_measurement_qubits
ValueError: 0 is not in list
```

because `_get_partial_measurement_qubits_from_tags` assumes the task tag `partial_measurement_qubits=0/1/2/3/4/5/6/7/8/9` is a subset of the result's `measured_qubits`, and raises on `list.index` when it is not.

## Fix

**`pad_measurements` (`qbraid/programs/gate_model/braket.py`)** — before adding padded measures, detect three failure modes:

1. **Internal collision** — two existing measures share `_target_index`.
2. **Padding collision** — an existing `_target_index` lies in `[k, k+m-1]`, the range Braket's counter will assign to padded measures.
3. **Out-of-range** — an existing `_target_index >= qubit_count`, which breaks Braket's `bit[qubit_count] b` emission.

When any of the three is present, rebase existing measures to sequential slots `[0..k-1]` in instruction order. Circuits with unique, in-range, non-overlapping labels are left untouched so user-chosen bit labels round-trip unchanged (e.g. `b[5] = measure q[0]` stays `b[5] = measure q[0]`).

**`_get_partial_measurement_qubits_from_tags` (`qbraid/runtime/aws/job.py`)** — when the tag's qubits are not all present in `bk_result.measured_qubits`, log a warning and return `None` (skipping partial-measurement filtering) instead of crashing. This lets users still load and inspect results from tasks that were submitted before this fix landed.

## Tests

Five new tests cover the edge cases:

- `test_pad_measurements_rebases_multi_register_collision` — reproduces the original bug via colliding `_target_index` values.
- `test_pad_measurements_rebases_when_padding_would_collide` — existing label lands in the counter's padded range.
- `test_pad_measurements_rebases_out_of_range_target_index` — existing label ≥ `qubit_count`.
- `test_pad_measurements_preserves_safe_user_labels` — unique, in-range, no-overlap labels are not rebased.
- `test_get_partial_measurement_qubits_from_tags_mismatch_returns_none` — exercises the `measured_qubits=[6..15]` vs tag `0..9` shape from the real failing task and asserts graceful `None` + warning instead of `ValueError`.

Four of the five fail on `main`; the preservation test passes both before and after (it's the "do nothing" path).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed measurement padding logic to detect and mitigate classical-bit collisions and index conflicts, ensuring valid quantum assembly output with properly sequential measurement indices.
  * Fixed partial measurement result processing to gracefully handle missing or inconsistent measurement tag data by returning appropriate null values and emitting diagnostic warnings instead of crashing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,8 @@ Types of changes:
 ### Fixed
 - Fixed pyqpanda3-to-QASM2 conversion emitting invalid `creg c[0]` declarations, which caused downstream parsers to reject the output and broke round-trip conversions (e.g. `cirq → pyqpanda3 → cirq`)
 - Fixed azure-quantum version mismatch in development requirements to align with package optional dependency constraints ([#1135](https://github.com/qBraid/qBraid/pull/1135))
+- Fixed classical bit collisions in Braket `pad_measurements` method by detecting internal collisions, padding collisions, and out-of-range indices; rebases existing measures to sequential indices when necessary to ensure valid QASM output ([#1160](https://github.com/qBraid/qBraid/pull/1160))
+- Fixed `BraketQuantumTask._get_partial_measurement_qubits_from_tags` to return `None` and log warning when tag qubits are missing from result measurements, preventing crashes during result processing ([#1160](https://github.com/qBraid/qBraid/pull/1160))
 
 ### Dependencies
 - Updated `azure-quantum` optional dependency from `>=2.0,<2.3` to `>=3.6.0,<4.0`; removed `azure-identity` from the `azure` extra ([#1125](https://github.com/qBraid/qBraid/pull/1125))
--- a/qbraid/programs/gate_model/braket.py
+++ b/qbraid/programs/gate_model/braket.py
@@ -151,10 +151,14 @@ class BraketCircuit(GateModelProgram):
         all qubits and no padding is applied.
         """
         # Track qubits that already have measurement instructions
-        partial_measurement_qubits: list[int] = []
-        for instruction in self._program.instructions:
-            if isinstance(instruction.operator, Measure):
-                partial_measurement_qubits.append(int(instruction.target[0]))
+        existing_measures = [
+            instruction
+            for instruction in self._program.instructions
+            if isinstance(instruction.operator, Measure)
+        ]
+        partial_measurement_qubits: list[int] = [
+            int(instruction.target[0]) for instruction in existing_measures
+        ]
 
         # Only apply padding when there is partial measurement or there is non-continguous qubits
         if (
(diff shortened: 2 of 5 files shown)


ryanhill1:
@yitchen-tim — this builds on top of the partial-measurement tagging you introduced in #1028, and touches the same `pad_measurements` path plus `_get_partial_measurement_qubits_from_tags`. Flagging you in case any of the rebase-on-collision behavior would conflict with how you use partial measurements in practice. In particular:

- `pad_measurements` now rebases existing measures' `_target_index` to sequential `[0..k-1]` when it detects an internal collision, a collision with the padded-counter range, or an out-of-range index. Circuits whose existing labels are unique, in-range, and don't overlap with the padded range are left untouched.
- `_get_partial_measurement_qubits_from_tags` now logs a warning and returns `None` (instead of raising `ValueError`) when the tag's qubits aren't all present in `bk_result.measured_qubits`. Does anything downstream in your flow rely on the previous exception behavior?

Does either change break anything for your use case?

yitchen-tim:
@ryanhill1 thanks for checking! The PR itself does not conflict with existing use cases. This PR is a cool work around for using multiple registers on Amazon Braket.

Adding some context: Amazon Braket service today natively supports only 1 register. The main reason is limited hardware support for mid-circuit measurements, therefore there is little need for more than 1 register. This is something that will change (we will support more than 1 register).

PR #1132, opened 2026-03-03
chore: replace qiskit-qir with qbraid-qir[qiskit]

eidrynbot:
## Summary

Replaces the archived `qiskit-qir` package with `qbraid-qir[qiskit]` for Qiskit to QIR conversions.

## Changes

- Updated `qiskit_extras.py` to use `qbraid_qir.qiskit.qiskit_to_qir` instead of `qiskit_qir.to_qir_module`
- Removed `qiskit-qir` from `requirements-dev.txt`
- Added `qbraid-qir[qiskit]>=0.6.0` dependency

## Dependencies

**This PR depends on [qBraid/qbraid-qir#271](https://github.com/qBraid/qbraid-qir/pull/271) being merged and released first.**

The qbraid-qir PR adds the `qbraid_qir.qiskit` module that this PR depends on.

## Benefits

- Removes dependency on archived package (`microsoft/qiskit-qir`)
- Provides Qiskit 2.x compatibility
- Consolidates QIR conversion code under qbraid-qir

## Testing

Tested locally with the qbraid-qir branch installed. The `qiskit_to_pyqir` conversion function works correctly with the new implementation.

PR #1163, opened 2026-04-20
Add Quantinuum NEXUS provider

vprusso:
## Summary

Adds a native Quantinuum provider to the qBraid runtime, following the same pattern as the recently merged OriginQ provider (#1148). This ports the Quantinuum integration originally maintained in [metriq-gym](https://github.com/unitaryfoundation/metriq-gym) (see unitaryfoundation/metriq-gym#731).

## What's included

- `qbraid/runtime/quantinuum/` — provider, device, job, and exceptions
- `qbraid/runtime/quantinuum/provider.py` — \`QuantinuumProvider\` with \`get_device\`/\`get_devices\` via qnexus
- `qbraid/runtime/quantinuum/device.py` — \`QuantinuumDevice\` with device status, pytket/qiskit transform, and full compile+execute submit pipeline
- `qbraid/runtime/quantinuum/job.py` — \`QuantinuumJob\` with status mapping, cancel, wall-clock execution time, and dlo-ordered counts retrieval
- 36 unit tests covering status mapping, provider discovery, device status/transform, job lifecycle (status/cancel/execution_time/result)
- Entry points and lazy-loading registrations
- New `quantinuum` optional dependency group: \`qnexus\`, \`pytket\`, \`pytket-qiskit\`

## Notes

- Authentication is handled implicitly by qnexus (\`qnx login\`).
- \`submit()\` currently blocks on \`qnx.jobs.wait_for(compile_job)\` during dispatch, consistent with the existing metriq-gym implementation. Compilation-as-async could be a follow-up.
- Counts are returned in \`BasisOrder.dlo\` (most-significant-bit-first) for consistency with other qBraid providers; pytket's native LSB-first ordering is converted at result time.
- The \`_map_quantinuum_status\` helper maps qnexus \`last_status\` strings to \`JobStatus\`, including an explicit \`CANCELLED\` state that metriq-gym's original implementation did not handle.

## Testing

Ran \`tests/runtime/quantinuum/\` with qnexus/pytket installed — 36/36 pass. Pylint 10.00/10, black and isort clean. End-to-end verification against real Quantinuum hardware will follow via the companion metriq-gym PR that points at this branch.

## Summary by CodeRabbit

* **New Features**
  * Added Quantinuum NEXUS provider integration enabling device discovery, quantum circuit submission (supporting pytket and Qiskit formats), and comprehensive job management including status tracking, cancellation, and result retrieval with measurement counts.


qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,183 @@
...
+
+            if isinstance(run_input, QuantumCircuit):
+                return [qiskit_to_tk(run_input)]
+        except ImportError:
+            pass
+
+        raise TypeError(
+            f"Unsupported run_input type {type(run_input)}; "
+            "expected pytket.Circuit or qiskit.QuantumCircuit"
+        )
```
ryanhill1: Add `pytket_to_qiskit` as a new conversion function in [`pytket_extras.py`](https://github.com/qBraid/qBraid/blob/c75eb4011de7142273dd883bbd8cafe90ea21d4d/qbraid/transpiler/conversions/pytket/pytket_extras.py) and then this step will be handled automatically by the transpiler.

qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,183 @@
...
+
+    def run(self, run_input, *args, **kwargs):
+        """Run a quantum program on this device.
+
+        Overridden to avoid the base class iterating single pytket circuits
+        as sequences; batching is handled inside :meth:`submit`.
+        """
+        return self.submit(run_input, *args, **kwargs)
```
ryanhill1: Not needed, can be removed.

qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,183 @@
...
+
+    # pylint: disable-next=arguments-differ
+    def submit(
+        self,
+        run_input,
+        shots: int | None = None,
+        **_: Any,
```
ryanhill1: can remove this extra `**` argument if it unused

qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,183 @@
...
+        compilation step to finish, then returns a :class:`QuantinuumJob`
+        referencing the asynchronous execution job.
+        """
+        # pylint: disable-next=import-outside-toplevel
+        import qnexus as qnx
+
+        # pylint: disable-next=import-outside-toplevel
+        from qnexus.models.language import Language
+
+        circuits_list = self.transform(run_input)
```
ryanhill1: This line can be removed. Handled automatically in parent class run method.

qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+        # NOTE: blocking wait during dispatch; compilation time depends on queue and program size.
+        logger.info(
+            "Waiting for Quantinuum compilation job %s to complete...",
+            getattr(compile_job, "id", getattr(compile_job, "job_id", str(compile_job))),
+        )
+        qnx.jobs.wait_for(compile_job)
+        compiled_refs = [item.get_output() for item in qnx.jobs.results(compile_job)]
+
+        nshots = int(shots or 1000)
```
ryanhill1: if we're using a fallback we should add it to the function signature instead of None to be clear. Or just require shots of type int

qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,183 @@
...
+            n_shots=[nshots] * len(compiled_refs),
+            backend_config=backend_config,
+            project=project,
+            language=Language.QIR,
+        )
+        job_id = (
+            getattr(execute_job, "id", None)
+            or getattr(execute_job, "job_id", None)
+            or str(execute_job)
+        )
```
ryanhill1: Can you look into `execute_job` and determine which of these 3 is the correct convention to get the job id? This should be straightword without fallbacks

qbraid/runtime/quantinuum/job.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class QuantinuumJob(QuantumJob):
+    """Quantinuum NEXUS job class."""
+
+    def __init__(
+        self,
+        job_id: str,
+        device: QuantinuumDevice | None = None,
+        job: Any | None = None,
```
ryanhill1: This should have a type

qbraid/runtime/quantinuum/job.py
```diff
@@ -0,0 +1,191 @@
...
+        # Quantinuum / pytket use least-significant-bit-first ordering by default.
+        # Convert to most-significant-bit-first (dlo = descending lexicographic order)
+        # for consistency with other qBraid providers.
+        all_counts: list[dict[str, int]] = []
+        for result_item in results:
+            counts = result_item.download_result().get_counts(basis=BasisOrder.dlo)
+            all_counts.append({"".join(map(str, k)): v for k, v in counts.items()})
+
+        measurement_counts = all_counts[0] if len(all_counts) == 1 else all_counts
+        device_id = self._device.id if self._device else "quantinuum"
```
ryanhill1: Does a Quantinuum job contain any details about which device it was executed on? If not this is ok. But if so, we can try to determine the device from the job metadata.

qbraid/runtime/quantinuum/provider.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @cached_method
+    def get_device(self, device_id: str) -> QuantinuumDevice:
+        """Get a specific Quantinuum device."""
+        device_id = device_id.strip()
+        backend_info = _get_backend_info(device_id)
+        return QuantinuumDevice(
+            profile=_build_profile(device_id, backend_info),
+            backend_info=backend_info,
+        )
```
ryanhill1: we can include backend_info as kwarg to `TargetProfile` (accepts extras), which would keep the consistency in the `QuantumDevice` interface. Could then be accessed via `device.profile.backend_info`

qbraid/runtime/quantinuum/provider.py
```diff
@@ -0,0 +1,106 @@
...
+
+    @cached_method
+    def get_devices(self, **kwargs) -> list[QuantinuumDevice]:  # pylint: disable=unused-argument
```
ryanhill1: can remove `**kwargs` if not used, and just surpress arguments differ warning

qbraid/runtime/quantinuum/provider.py
```diff
@@ -0,0 +1,106 @@
...
+        ``kwargs`` is accepted for interface compatibility with the base
+        :class:`~qbraid.runtime.QuantumProvider.get_devices` signature; the
+        NEXUS API does not currently expose additional filtering parameters.
+        """
+        # pylint: disable-next=import-outside-toplevel
+        import qnexus as qnx
+
+        df = qnx.devices.get_all(issuers=[qnx.devices.IssuerEnum.QUANTINUUM]).df()
+        device_ids = [
+            str(row.get("device_name") or row.get("name") or "").strip() for _, row in df.iterrows()
```
ryanhill1: Can you determine correct way to get device name and use it directly?

pyproject.toml
```diff
@@ -49,6 +49,7 @@ cudaq = ["cudaq>=0.9.0,<0.14.0"]
 ionq = ["qiskit-ionq>=0.5.12"]
 oqc = ["oqc-qcaas-client>=3.11.0"]
 origin = ["pyqpanda3"]
+quantinuum = ["qnexus>=0.39.0", "pytket>=1.31", "pytket-qiskit>=0.71.0"]
```
ryanhill1: pytket-qiskit would not technically be required to use this provider correct? The pytket-qiskit extra just allows you submitting in qiskit format in addition to pytket?

If so, there are many extras that fall into this category. So might be worth only including what is minimally required

ryanhill1 (changes requested):

qbraid/runtime/quantinuum/device.py
vprusso: Done — added `pytket_to_qiskit` in `qbraid/transpiler/conversions/pytket/pytket_extras.py` (plus the accompanying `__init__` export). Removed the manual `qiskit_to_tk` conversion from `transform()` entirely; the transpiler now handles `qiskit → qasm2 → pytket` automatically via the existing `qiskit_to_qasm2` / `qasm2_to_pytket` edges, so the whole custom `transform()` method was dropped.

qbraid/runtime/quantinuum/device.py
vprusso: Removed — the `run()` override is gone. The base-class pipeline handles single-vs-list and transpilation, and `submit()` now accepts `Circuit | list[Circuit]` directly.

qbraid/runtime/quantinuum/device.py
vprusso: Removed — no `**kwargs` catch-all on `submit()` anymore.

qbraid/runtime/quantinuum/device.py
vprusso: Removed — `submit()` takes `run_input` as-is (`Circuit` or `list[Circuit]`), normalizes to a list locally, and uploads directly. No intra-`submit` `self.transform()` call.

qbraid/runtime/quantinuum/device.py
vprusso: Changed to `shots: int = 1000` — default is visible in the signature, no `None` fallback.

qbraid/runtime/quantinuum/device.py
vprusso: Confirmed via `qnexus.models.references.ExecuteJobRef` — it's a pydantic model with a single `id: uuid.UUID` field. Replaced the 3-way fallback with `str(execute_job.id)`.

qbraid/runtime/quantinuum/job.py
vprusso: Typed as `ExecuteJobRef | None` (from `qnexus.models.references`), imported under `TYPE_CHECKING`. `_get_ref()` now returns `ExecuteJobRef` as well.

qbraid/runtime/quantinuum/job.py
vprusso: Yes — `ExecuteJobRef.backend_config_store` is a `QuantinuumConfig` whose `device_name` field holds the target device. Added a `_resolve_device_id()` helper that prefers `self._device.id`, then `ref.backend_config_store.device_name`, and finally falls back to the generic `"quantinuum"` label. Covered by two new tests (`test_result_derives_device_id_from_job_metadata`, `test_result_falls_back_to_generic_device_id`).

qbraid/runtime/quantinuum/provider.py
vprusso: Done — `backend_info` is now passed as an extra kwarg to `TargetProfile` (which has `extra="allow"`). `QuantinuumDevice.__init__` no longer takes `backend_info` separately; `device.backend_info` now reads from `self.profile.backend_info`, keeping the `QuantumDevice` constructor signature consistent across providers.

qbraid/runtime/quantinuum/provider.py
vprusso: Removed `**kwargs` from `get_devices` and swapped the `arguments-differ` suppression onto the method directly.

qbraid/runtime/quantinuum/provider.py
vprusso: The canonical column is `device_name` — confirmed by the `_get_backend_info` helper in the same file, which queries `df["device_name"] == device_name`. Dropped the `name` fallback.

pyproject.toml
vprusso: Good point — dropped `pytket-qiskit` from the `quantinuum` extras. The new `pytket_to_qiskit` transpiler conversion is gated by `@requires_extras("pytket.extensions.braket")` style annotation, so qiskit support is opt-in via the `pytket`/`qiskit` extras rather than forced into the provider install.

qbraid/runtime/quantinuum/provider.py
```diff
@@ -0,0 +1,106 @@
...
+    matching_rows = df.loc[df["device_name"] == device_name]
+
+    if matching_rows.empty:
+        available_devices = df["device_name"].tolist()
+        raise ResourceNotFoundError(
+            f"Device '{device_name}' not found in Quantinuum device list. "
+            f"Available devices: {available_devices}"
+        )
+
+    return matching_rows.iloc[0]["backend_info"]
```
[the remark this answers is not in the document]
vprusso: Fixed in 4c2f88d8. Extracted `_fetch_quantinuum_devices_df()` as a single helper and rewrote `get_devices()` to build `QuantinuumDevice` instances directly from the fetched dataframe rows — no round-trip through `get_device` / `_get_backend_info`, so only one `qnx.devices.get_all` call per `get_devices` invocation. Also dropped the `name`/`device_name` fallback (per the maintainer's other comment); the canonical NEXUS column is `device_name`. Added `test_get_devices_makes_single_remote_call` as a regression guard for the N+1 pattern.

requirements-dev.txt
```diff
@@ -19,6 +19,7 @@ qiskit-aer>=0.15.0; python_version < "3.13"
 qiskit-qir
 qiskit-ionq>=0.5.12
 pyqpanda3>=0.2.0; python_version < "3.13"
+qnexus>=0.39.0
```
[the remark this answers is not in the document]
vprusso: Added `pytket-qiskit>=0.71.0` to `requirements-dev.txt`. Note: the `quantinuum` extra in `pyproject.toml` deliberately does *not* include `pytket-qiskit` (per the maintainer's separate comment — pytket-qiskit is only needed if users want to submit `qiskit.QuantumCircuit` inputs, not strictly required for the provider itself). The dev pin ensures CI can exercise the new `pytket_to_qiskit` transpiler conversion and any qiskit-bridge code paths.

tests/runtime/quantinuum/test_quantinuum_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+
+def _make_device(device_id: str = "H1-1E", simulator: bool = True):
+    """Helper to create a QuantinuumDevice with mocked dependencies."""
+    backend_info = _make_backend_info()
+    profile = MagicMock()
+    profile.device_id = device_id
+    profile.simulator = simulator
+    device = QuantinuumDevice(profile=profile, backend_info=backend_info)
+    # QuantumDevice.id is a property that returns profile.device_id
+    type(device).id = property(lambda self: self.profile.device_id)
```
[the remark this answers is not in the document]
vprusso: Done — removed the class-level `type(device).id = property(...)` mutation. Confirmed `QuantumDevice.id` (the base class property at `qbraid/runtime/device.py:86`) returns `self.profile.device_id`, so the `MagicMock` profile with `profile.device_id = device_id` is already enough to make `device.id` work correctly in tests.

qbraid/runtime/quantinuum/provider.py
```diff
@@ -0,0 +1,111 @@
...
+        :class:`QuantinuumDevice` from the returned dataframe rows, avoiding
+        the N+1 pattern that would otherwise re-fetch the device list inside
+        :meth:`get_device` for each entry.
+        """
+        df = _fetch_quantinuum_devices_df()
+        devices: list[QuantinuumDevice] = []
+        for _, row in df.iterrows():
+            device_id = str(row["device_name"]).strip()
+            if not device_id:
+                continue
```
ryanhill1: Is this `continue` line actually reachable? Does `qnx.devices.get_all(issuers=[qnx.devices.IssuerEnum.QUANTINUUM]).df()` ever return a DataFrame in which one of the `row["device_name"]` is `None` or an empty string?

qbraid/runtime/quantinuum/device.py
```diff
@@ -0,0 +1,136 @@
...
+            return DeviceStatus.ONLINE
+        if status in (DeviceStateEnum.MAINTENANCE, DeviceStateEnum.RESERVED_MAINTENANCE):
+            return DeviceStatus.UNAVAILABLE
+        return DeviceStatus.OFFLINE
+
+    def submit(  # pylint: disable=arguments-differ
+        self,
+        run_input: Circuit | list[Circuit],
+        shots: int = 1000,
+    ) -> QuantinuumJob:
```
ryanhill1, line 82: I think it would improve the UX if we added `project_name` and `opt_level` as optional keyword arguments here, instead requiring that they're coded as environment variables.

qbraid/runtime/quantinuum/job.py
```diff
@@ -0,0 +1,209 @@
...
+
+        try:
+            ref = self._get_ref()
+            last_status = getattr(ref, "last_status", None)
+        except Exception as exc:
+            raise QuantinuumJobError(f"Unable to retrieve job status for {self.id}") from exc
+
+        status = _map_quantinuum_status(last_status)
+        if status == JobStatus.FAILED:
+            last_message = getattr(ref, "last_message", None)
```
ryanhill1:
```suggestion
            last_message = reff.last_message
```

From what I can find from the qnexus docs, the [`ExecuteJobRef`](https://docs.quantinuum.com/nexus/_modules/qnexus/models/references.html#ExecuteJobRef) is a pydantic schema that inherits from [`JobRef`](https://docs.quantinuum.com/nexus/_modules/qnexus/models/references.html#JobRef), which includes `last_message` as a required field of type `string`.

qbraid/runtime/quantinuum/job.py
```diff
@@ -0,0 +1,209 @@
...
+        """Resolve the target device name for a job.
+
+        Prefers the device this job was constructed with, then tries to read
+        the device name from the NEXUS job's ``backend_config_store`` (set to
+        a ``QuantinuumConfig`` when the job was dispatched via qBraid).
+        Falls back to the generic ``"quantinuum"`` label if neither is set.
+        """
+        if self._device is not None:
+            return self._device.id
+        backend_config = getattr(ref, "backend_config_store", None)
```
ryanhill1:
```suggestion
        backend_config = ref.backend_config
```

Looking at [`JobRef`](https://docs.quantinuum.com/nexus/_modules/qnexus/models/references.html#JobRef) I believe that it has property `backend_config` that returns a `BackendConfig`. The resolves from input field `backend_config_store` which could be `BackendConfig | None`.

qbraid/runtime/quantinuum/job.py
```diff
@@ -0,0 +1,209 @@
...
+        device_name = getattr(backend_config, "device_name", None)
+        if device_name:
+            return str(device_name)
+        return "quantinuum"
```
ryanhill1: It seems `BackendConfig.device_name` is only defined for the IBM devices: https://github.com/quantinuum-dev/quantinuum-schemas/blob/main/quantinuum_schemas/models/backend_config.py

qbraid/runtime/quantinuum/job.py
```diff
@@ -0,0 +1,209 @@
...
+        if self.status() != JobStatus.COMPLETED:
+            return None
+        ref = self._get_ref()
+        last_status_detail = getattr(ref, "last_status_detail", None)
+        if last_status_detail is None:
+            raise QuantinuumJobError(
+                f"Execution time not available for {self.id}: last_status_detail is missing"
+            )
+        completed_time = getattr(last_status_detail, "completed_time", None)
+        running_time = getattr(last_status_detail, "running_time", None)
```
ryanhill1: can you verify that each place `getattr` is used, it is because there really is a possability of `AttributeError`  accessing these fields?

CHANGELOG.md
ryanhill1, line 1: can you add an item about how the `pytket_to_qiskit` function was added, and add a test case for it if not already present

CHANGELOG.md
```diff
@@ -16,6 +16,28 @@ Types of changes:
 ## [Unreleased]
 
 ### Added
+- Added Quantinuum NEXUS provider integration (`qbraid.runtime.quantinuum`) with `QuantinuumProvider`, `QuantinuumDevice`, and `QuantinuumJob` classes. Supports single-circuit and batch submission via the NEXUS compile + execute pipeline; accepts both pytket `Circuit` and `qiskit.QuantumCircuit` inputs. Counts are returned in MSB-first (`BasisOrder.dlo`) ordering for consistency with other qBraid providers. ([#1163](https://github.com/qBraid/qBraid/pull/1163))
```
ryanhill1: supports every other program type that has a conversion path to `pytket` as well, probably don't need to call out `qiskit` specifically because it makes that sound like the only other accepted type.

ryanhill1 (changes requested):
Also please update [`docs/api/qbraid.runtime.rst`](https://github.com/vprusso/qBraid/blob/add-quantinuum-provider/docs/api/qbraid.runtime.rst) with `quantinuum` provider

qbraid/runtime/quantinuum/provider.py
vprusso: You're right — the NEXUS `device_name` column is always a non-empty string. Removed the `if not device_id: continue` guard and collapsed the loop into a list comprehension (`provider.py` L104-110).

qbraid/runtime/quantinuum/device.py
vprusso: Added `project_name` and `optimisation_level` as optional kwargs to `submit()`, with env-var fallbacks (`QUANTINUUM_NEXUS_PROJECT_NAME` / `QUANTINUUM_NEXUS_OPT_LEVEL`) preserved so existing dispatchers don't break. Docstring documents both. Since `run()` forwards `*args`/`**kwargs` to `submit()`, users can pass these as `device.run(circuit, shots=N, project_name="my-proj", optimisation_level=2)`.

qbraid/runtime/quantinuum/job.py
vprusso: Replaced `getattr(ref, "last_message", None)` with direct `ref.last_message` access in both `status()` and `result()` — confirmed it's a required `str` field on `JobRef`.

qbraid/runtime/quantinuum/job.py
vprusso: Done — `_resolve_device_id` now uses `ref.backend_config` (the resolving property on `JobRef` that falls back to `_fetch_by_id` if `backend_config_store` is `None`).

qbraid/runtime/quantinuum/job.py
vprusso: Good catch — `device_name` only exists on `QuantinuumConfig` (and `IBMQConfig`), not the base `BackendConfig`. Changed to an explicit `isinstance(backend_config, QuantinuumConfig)` check in `_resolve_device_id` and return the field directly; falls back to `"quantinuum"` for any other config subclass. Updated `test_result_falls_back_to_generic_device_id` to exercise the non-Quantinuum path with `AerConfig`.

qbraid/runtime/quantinuum/job.py
vprusso: Audited all `getattr` calls and removed them where the attribute is guaranteed by the schema: `ref.last_status`, `ref.last_message`, `ref.last_status_detail`, `last_status_detail.completed_time`, and `last_status_detail.running_time` are now direct accesses (`JobRef` has the first three as required/typed, and `JobStatus` is a `NamedTuple` where the time fields are always present as `datetime | None`). The surrounding `try/except` around the ref lookup stays to catch remote-call errors. Test `test_status_wraps_ref_lookup_errors` added to cover that path.

CHANGELOG.md
vprusso: Added a dedicated CHANGELOG entry for `pytket_to_qiskit` (linked to this PR) and a unit test `test_bell_state_pytket_to_qiskit` in `tests/transpiler/pytket/test_from_pytket.py` that exercises both the direct `pytket_to_qiskit` call and the full `transpile(circuit, "qiskit")` path, guarded by `pytest.importorskip` on `qiskit` and `pytket.extensions.qiskit`.

CHANGELOG.md
vprusso: Rewrote to "accepts any program type reachable to pytket via the qBraid transpiler graph" — no longer singles out qiskit.

vprusso:
Pushed c247391d addressing the full review:

- **`docs/api/qbraid.runtime.rst`**: added `quantinuum` to the submodules list
- **`CHANGELOG.md`**: broadened the Quantinuum entry to reference "any program type reachable to pytket via the transpiler graph" instead of calling out qiskit specifically; added a separate entry for the new `pytket_to_qiskit` transpiler conversion
- **`qbraid/runtime/quantinuum/provider.py`**: dropped the unreachable empty-`device_name` guard, collapsed `get_devices` into a list comprehension
- **`qbraid/runtime/quantinuum/device.py`**: `submit()` now takes `project_name` and `optimisation_level` as optional kwargs with env-var fallbacks
- **`qbraid/runtime/quantinuum/job.py`**: removed all unnecessary `getattr` calls, using direct attribute access on `JobRef` / `JobStatus` fields; `_resolve_device_id` now uses `ref.backend_config` and an explicit `isinstance(..., QuantinuumConfig)` check
- **`tests/transpiler/pytket/test_from_pytket.py`**: added `test_bell_state_pytket_to_qiskit` covering both the direct conversion and the `transpile(...)` path

37 Quantinuum tests + 1 new pytket-to-qiskit test all pass. Pylint 10/10, black + isort clean.

requirements-dev.txt
```diff
@@ -19,6 +19,8 @@ qiskit-aer>=0.15.0; python_version < "3.13"
...
```
ryanhill1, line 22: super small nitpick: can you move this down under "optional runtime dependencies"

ryanhill1 (commented):
You may need to add `qnexus` to `autodoc_mock_imports` in order for docs build to pass: https://github.com/vprusso/qBraid/blob/add-quantinuum-provider/docs/conf.py#L48

As for the dependency conflict in the unit tests, I think it is because `pytket-qiskit` requires "qiskit>=2.3" while `qiskit-qir` requires `qiskit>=1.0.0,<2.0`. I have a draft PR that replaces the deprecated `qiskit-qir` package with a `qbraid-qir[qiskit]` extra: 
- https://github.com/qBraid/qBraid/pull/1132
- https://github.com/qBraid/qbraid-qir/pull/271

that should resolve this and unlock the latest version of qiskit. I can try to get that merged tomorrow or early next week.

In the meantime, we can maybe comment-out the `pytket-qiskit` dependency in `requirements-dev.txt` so that the conversion test is skipped for now so that the tests pass and we can merge this, and add the requirement back once the qbraid-qir[qiskit] update is made.

vprusso:
Pushed [a7535ecb](https://github.com/qBraid/qBraid/pull/1163/commits/a7535ecb) addressing both:

1. **docs build**: added `qnexus` and `quantinuum_schemas` to `autodoc_mock_imports` in `docs/conf.py` so the Sphinx build no longer needs those packages installed.
2. **dep conflict**: commented out `pytket-qiskit` in `requirements-dev.txt` with a comment linking to #1132. The new `test_bell_state_pytket_to_qiskit` test uses `pytest.importorskip("pytket.extensions.qiskit")` so it skips cleanly without the dep (verified locally).

Happy to uncomment the dev dep in a follow-up once #1132 / qbraid-qir[qiskit] lands.

ryanhill1 (changes requested):
Last thing: can you add tests to satisfy codecov?

It looks like there aren't yet tests for:
- `QuantinuumDevice.submit()`
- `QuantinuumJob._get_ref()` where `self._job is None`
- Exception handling in `QuantinuumJob.result()`
- `QuantinuumProvider.__hash__()`

I don't have a Quantinuum Nexus API login, so these will be important as I can't test things end-to-end myself.

vprusso:
Pushed [398861ae](https://github.com/qBraid/qBraid/pull/1163/commits/398861ae) adding the 4 missing coverage areas:

- **`QuantinuumDevice.submit()`**: `test_submit_runs_full_compile_execute_pipeline` exercises the upload → compile → execute flow with mocked `qnx` calls; `test_submit_accepts_project_and_opt_level_kwargs` covers the kwarg path for `project_name` / `optimisation_level`
- **`QuantinuumJob._get_ref()` with `self._job is None`**: `test_get_ref_lazily_fetches_when_job_is_none` (verifies `qnx.jobs.get(id=...)` is called once and cached) and `test_get_ref_wraps_remote_errors`
- **Exception handling in `QuantinuumJob.result()`**: `test_result_wraps_remote_errors` covers the `qnx.jobs.results` failure path
- **`QuantinuumProvider.__hash__()`**: `test_provider_is_hashable` confirms instances are hashable, each instance has a distinct hash, and they can be used as dict keys / set members

Total now 43 tests (from 37), and `pytest-cov` locally reports **100%** coverage on `qbraid.runtime.quantinuum`. Pylint 10/10, black + isort clean.

ryanhill1 (approved):

PR #1170, opened 2026-04-29
Fix barrier removal for QIR simulator devices

TheGupta2012:
**Fix** — Bug Fix

Adds barrier removal to `QbraidDevice.transform()` so QASM programs containing `barrier` instructions can be submitted to QIR-based simulator devices without triggering linker errors (`Failed to link some declared functions: __quantum__qis__barrier__body`). Barriers are semantically no-ops in quantum computing but are unsupported by the qBraid QIR runtime.

## Changes

- **`_QIR_DEVICE_IDS`**: Added module-level set in `qbraid/runtime/native/device.py` enumerating device IDs that require barrier stripping (initially `{"qbraid:qbraid:sim:qir-sv"}`).
- **`QbraidDevice.transform()`**: New method that intercepts QASM run input for QIR devices, uses `load_program()` + `pyqasm` to detect and remove barrier operations before submission; non-QASM or non-QIR inputs pass through unchanged.
- **Unit tests**: Added 8 tests in `tests/runtime/native/test_qir_simulator_remote.py` covering barrier removal, no-barrier passthrough, non-QIR device preservation, non-QASM passthrough, multiple barriers, single-qubit barrier, barrier between gates/measurements, and barrier-only circuits.


ryanhill1:
not needed

PR #1187, opened 2026-05-12
Add batch job support (as_batch=True)

TheGupta2012:
## Summary

- Add `as_batch=True` parameter to `device.run()` / `device.submit()` to submit a list of circuits as a single batch job (one VRN, one status, one result)
- `QbraidJob.result()` returns `list[Result]` for batch jobs (`numCircuits > 1`) and a single `Result` for regular jobs
- Batch jobs are limited to 2000 circuits max
- Compatible with existing `GroupJobSession` context

### DIFF
--- a/qbraid/runtime/native/device.py
+++ b/qbraid/runtime/native/device.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# pylint: disable=arguments-differ
+# pylint: disable=arguments-differ,too-many-arguments
 
 """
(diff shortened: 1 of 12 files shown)


qbraid/runtime/native/device.py
TheGupta2012, line 192: @ryanhill1 I guess we're wrapping the run_input in a `list[list[Program]]` in case of a batch and is kind of non-obvious from the first glance - will add a comment specifying this

ryanhill1 (approved):

PR #1211, opened 2026-06-05
Qrisp as a supported framework

### LINKED ISSUE #1186: [FEATURE] Add new conversion(s) of your choice (1/2)

ACE07-Sev:
## Summary of changes

Closes #1186 .

- Added `qrisp` as a supported framework to converters.
- Added `qrisp` to supported programs via `QrispCircuit`.

New paths
```
`qrisp` -> `cirq`
`qrisp` -> `qiskit`
`qrisp` -> `pytket`
`qrisp` -> `qasm2`
`qrisp` -> `qasm3`
`qiskit` -> `qrisp`
```

Currently have added `test_coverage_from_qrisp.py`, and `test_program_circuits_qrisp.py`, but need guidance for additional testers based on what you'd like me to add.

This is WIP since all tests pass but on CI for some reason `test_coverage_from_qrisp.py` seems to have issues with three gates. I tried to reproduce this exactly locally but no luck, it passes locally. Additionally, `qiskit_to_qrisp` seems to affect unrelated conversion paths (regardless of `weight` being lower than other `qiskit_to_x` converters), so need to fix that as well to allow for a way into `qrisp` from other frameworks.


ACE07-Sev:
@ryanhill1 if you can kindly try the `test_coverage_from_qrisp.py` locally to see if it also passes on your side or not would be immensely appreciated.

requirements-dev.txt
```diff
@@ -5,6 +5,7 @@ cudaq>=0.9.0,<0.14.0; python_version < "3.13"
 pennylane>=0.43; python_version >= "3.11"
 pytket>=1.31
 qiskit>=1.0,<3.0
+qrisp>=0.7
```
ryanhill1, line 12: This is almost certainly the CI-vs-local discrepancy. `qrisp>=0.7` lets CI (Python 3.10) install 0.7.x, which lacks `to_cirq()/to_qiskit()/to_pytket()` — the program tests then fail with `AttributeError: ... has no attribute 'to_cirq'`. qrisp 0.8 requires Python >=3.11, so pin it to match `pyproject.toml` and add the marker:

```suggestion
qrisp>=0.8; python_version >= "3.11"
```

pyproject.toml
```diff
@@ -58,6 +58,7 @@ pyquil = ["pyquil>=5.0.0rc3", "qcs-sdk-python>=0.26.1"]
 pytket = ["pytket>=1.31"]
 qir = ["qbraid-qir>=0.2.0,<=0.5.1"]
 qiskit = ["qiskit>=1.0,<3.0", "qiskit-ibm-runtime>=0.39.0,<0.42", "qiskit-qasm3-import>=0.5.1", "packaging>=20.0"]
+qrisp = ["qrisp>=0.8"]
 rigetti = ["pyquil>=5.0.0rc3; python_version < '3.13'", "qcs-sdk-python>=0.26.1; python_version < '3.13'"]
```
ryanhill1, line 63: `qrisp>=0.8` is fine, but note it requires Python >=3.11 while qBraid is `requires-python = ">=3.10"`, so `pip install qbraid[qrisp]` is unresolvable on 3.10. Consider adding a marker for robustness so the extra is silently skipped there:

```suggestion
qrisp = ["qrisp>=0.8; python_version >= '3.11'"]
```
Either way, `requirements-dev.txt` needs the same constraint (see that comment).

README.md
```diff
@@ -114,7 +114,8 @@ Below, `QPROGRAM_REGISTRY` maps shorthand identifiers for supported quantum prog
  'stim': stim._stim_sse2.Circuit,
  'pulser': pulser.sequence.sequence.Sequence,
  'pyqpanda3': pyqpanda3.core.QProg,
- 'autoqasm': autoqasm.program.program.Program}
+ 'autoqasm': autoqasm.program.program.Program
+ 'qrisp': qrisp.circuit.circuit.QuantumCircuit}
 ```
```
ryanhill1: Missing comma after the `autoqasm` entry — the displayed dict is syntactically invalid.

```suggestion
 'autoqasm': autoqasm.program.program.Program,
 'qrisp': qrisp.circuit.circuit.QuantumCircuit}
```

qbraid/programs/gate_model/qrisp.py
```diff
@@ -0,0 +1,87 @@
...
+        idle_qubits = [q for q, d in qubit_depths.items() if d == 0]
+        num_active_qubits = self.program.num_qubits() - len(idle_qubits)
+        new_program = qrisp.QuantumCircuit(num_active_qubits, self.num_clbits)
+
+        for instr in self.program.data:
+            qubits = [self.program.qubits.index(q) for q in instr.qubits if q not in idle_qubits]
+            clbits = [self.program.clbits.index(c) for c in instr.clbits]
+            new_program.append(instr.op, qubits, clbits)
+
+        self.program = new_program
```
ryanhill1, line 84: Bug: this remaps each gate's qubits using **old** indices (`self.program.qubits.index(q)`) into a circuit that has fewer qubits, so when an idle qubit is *not* trailing it raises `Tried to adress qubit with index N in a circuit with M qubits`. Repro:

```python
circ = QuantumCircuit(3); circ.h(0); circ.cx(0, 2)  # qubit 1 idle
QrispCircuit(circ).remove_idle_qubits()  # raises
```

The existing test only uses a trailing idle qubit, so it doesn't catch this. Idle qubits have depth 0 (no gate touches them), so the `if q not in idle_qubits` filter is a no-op — the real fix is to remap to the new contiguous indices:

```suggestion
        index_map = {
            q: i
            for i, q in enumerate(
                q for q in self.program.qubits if q not in idle_qubits
            )
        }

        for instr in self.program.data:
            qubits = [index_map[q] for q in instr.qubits]
            clbits = [self.program.clbits.index(c) for c in instr.clbits]
            new_program.append(instr.op, qubits, clbits)

        self.program = new_program
```
Worth adding a test with a non-trailing idle qubit too.

qbraid/transpiler/conversions/qiskit/__init__.py
```diff
@@ -28,16 +28,19 @@
...
 from .qiskit_to_qasm3 import qiskit_to_qasm3
+# from .qiskit_to_qrisp import qiskit_to_qrisp
 
 __all__ = [
     "qiskit_to_qasm2",
     "qiskit_to_qasm3",
     "qiskit_to_braket",
     "qiskit_to_pyqir",
     "qiskit_to_ionq",
+    "qiskit_to_qrisp",
```
ryanhill1, line 46: Since the import is commented out (line 37) and `qiskit_to_qrisp` is also in the autosummary (line 31), this `__all__` entry is dangling. It doesn't raise only because a submodule of the same name exists, so the name resolves to the **module object** rather than the function — `from ...qiskit import qiskit_to_qrisp` silently returns a module and Sphinx autosummary will misdocument/fail. While the converter is disabled, please drop it from `__all__` and the docstring autosummary (and ideally remove the commented import rather than leave it).

qbraid/transpiler/conversions/qrisp/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+   :toctree: ../stubs/
+
+   qrisp_to_cirq
+   qrisp_to_pytket
+   qrisp_to_qiskit
+
+"""
+from .qrisp_to_cirq import qrisp_to_cirq
+from .qrisp_to_pytket import qrisp_to_pytket
+from .qrisp_to_qiskit import qrisp_to_qiskit
```
ryanhill1, line 38: `qrisp_to_qasm2.py` and `qrisp_to_qasm3.py` are added in this PR but never imported here, so the conversion-discovery scan (`dir(submodule)` + `inspect.isfunction`) never registers them — verified that only `qrisp_to_cirq/pytket/qiskit` end up in `conversion_functions`. Either wire them up (add the imports, `__all__`, and autosummary entries) or remove the two files.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,189 @@
...
+
+
+def convert_from_qrisp_to_x(target, circuit_name, circuits, graph):
+    """Construct a Qrisp circuit with the given gate, transpile it to
+    target program type, and check equivalence.
+    """
+    source_circuit = circuits[circuit_name]
+    circuit = transpile(source_circuit, "cirq", conversion_graph=graph)
+    qasm = circuit.to_qasm()
+    cirq_circuit = circuit_from_qasm(qasm)
```
ryanhill1: This `circuit.to_qasm()` is where the coverage test fails even on py3.11 + qrisp 0.8.2: for `CPGate`, `MCRXGate`, `MCXGate` cirq raises `ValueError: Invalid format specifier 'half_turns' for object of type 'float'`, giving ~89% vs the asserted `1.0` baseline (line 149). So `qrisp→cirq` itself succeeds; the round-trip through cirq's QASM2 exporter is the weak link. Options: route around cirq's `to_qasm()` for these gates, special-case them, or lower/justify the baseline so the test reflects reality.

qbraid/transpiler/conversions/qrisp/qrisp_to_cirq.py
```diff
@@ -0,0 +1,47 @@
...
+"""
+Module defining Qrisp conversions
+
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from cirq import Circuit
```
ryanhill1: Nit: this eagerly imports cirq at module top, while the sibling converters (`qrisp_to_qiskit/pytket/...`) lazy-load via `LazyLoader`. Harmless since cirq is a native dep, but inconsistent — consider matching the lazy pattern.

ryanhill1 (changes requested):
Thanks for this! I checked out the branch and ran the new tests in two environments to chase down the CI-vs-local discrepancy: **py3.10 + qrisp 0.7.17** (what `requirements-dev.txt` actually resolves to) and **py3.11 + qrisp 0.8.2** (what `pyproject.toml` targets). Summary of findings below; details are in the inline comments.

### Blocking
1. **Version-pin mismatch is the likely CI-vs-local cause.** `pyproject.toml` pins `qrisp>=0.8` but `requirements-dev.txt` allows `qrisp>=0.7`. The converters rely on `to_cirq()/to_qiskit()/to_pytket()/to_qasm2/3()`, which **only exist in qrisp ≥0.8**. On 0.7.x the program tests fail immediately with `AttributeError: 'QuantumCircuit' object has no attribute 'to_cirq'`. Since qrisp ≥0.8 requires Python ≥3.11, CI on 3.10 pulls 0.7.x and breaks.
2. **The coverage test fails even on py3.11 + qrisp 0.8.2.** It asserts 100% coverage but gets ~89% — `CPGate`, `MCRXGate`, `MCXGate` fail. The failure is `ValueError: Invalid format specifier 'half_turns'` raised by cirq's `circuit.to_qasm()` on controlled/multi-controlled gates, i.e. the test's round-trip path, not the qrisp converter itself.
3. **Correctness bug in `remove_idle_qubits`** — it only works when idle qubits are trailing; a middle idle qubit raises `Tried to adress qubit with index 2 in a circuit with 2 qubits`. The unit test only covers the trailing case, so it's masked.

### Should fix
4. `qrisp_to_qasm2.py` / `qrisp_to_qasm3.py` exist but aren't imported in the `qrisp` conversions `__init__`, so they're never registered (verified — only cirq/pytket/qiskit register).
5. `qiskit/__init__.py` keeps `qiskit_to_qrisp` in `__all__` and the autosummary while the import is commented out.

### Verified good
- The 4 `test_program_circuits_qrisp.py` tests pass on py3.11 + qrisp 0.8.
- Graceful degradation on Python 3.10 works: `_dynamic_importer` swallows the missing-qrisp import, so the node/edges are simply skipped and `ConversionGraph` builds without crashing — provided `requirements-dev.txt` is pinned so qrisp 0.7 isn't installed on 3.10 (see comment #1).

qbraid/programs/gate_model/qrisp.py
ACE07-Sev: Ooh I totally missed this. Thanks for catching it. I was doing a hacky way before it and I only tested it on that (which worked, but super hacky) and I forgot to test this for the new one again hehe.

Fixed, thank you again!

tests/transpiler/qrisp/test_coverage_from_qrisp.py
ACE07-Sev: Weird thing is it's passing locally. Locally I get 1.0 baseline. I'll fix it for CI since now we know what's causing it, but need to find out why it passes locally.

qbraid/transpiler/conversions/qrisp/qrisp_to_cirq.py
```diff
@@ -0,0 +1,46 @@
...
+def qrisp_to_cirq(qrisp_qc: qrisp_.QuantumCircuit) -> Circuit:
+    """Returns a Cirq circuit equivalent to the input Qrisp circuit.
+
+    Args:
+        qrisp_qc: Qrisp circuit to convert to a Cirq circuit.
+
+    Returns:
+        Cirq Circuit object equivalent to the input Qrisp circuit.
+    """
+    return qrisp_qc.to_cirq()
```
ryanhill1: **Root cause of the coverage-test failures.** `qrisp_qc.to_cirq()` returns a correct cirq circuit, but its gate exponents are `numpy.longdouble`. cirq's QASM exporter (`QasmArgs.format_field`) only handles the `half_turns` spec when `isinstance(value, (float, int))` — `np.float64` subclasses `float`, but `np.longdouble` does **not** — so `circuit.to_qasm()` raises `ValueError: Invalid format specifier 'half_turns'` on `CPGate`/`MCRXGate`/`MCXGate`. Casting the exponents to plain `float` fixes all three (verified: `to_qasm` succeeds and the round-trip is `allclose=True`), which also lets the coverage baseline return to `1.0`:

```suggestion
    circuit = qrisp_qc.to_cirq()
    # qrisp emits extended-precision (numpy.longdouble) gate exponents, which cirq's QASM
    # exporter cannot format ("Invalid format specifier 'half_turns'"). Recast them to
    # plain Python float so downstream to_qasm() works.
    normalized = cirq.Circuit()
    for op in circuit.all_operations():
        gate = op.gate
        exponent = getattr(gate, "_exponent", None)
        if exponent is not None and not isinstance(exponent, (int, float)):
            global_shift = float(getattr(gate, "_global_shift", 0.0))
            try:
                gate = type(gate)(exponent=float(exponent), global_shift=global_shift)
            except TypeError:
                gate = type(gate)(exponent=float(exponent))
        normalized.append(gate.on(*op.qubits))
    return normalized
```

qbraid/transpiler/conversions/qiskit/qiskit_to_qrisp.py
```diff
@@ -0,0 +1,71 @@
...
+        "s",
+        "t",
+        "measure",
+        "global_phase",
+        "swap",
+    ]
+
+    target = Target.from_configuration(basis_gates=basis_set)
+
+    return convert_from_qiskit(transpile(qiskit_qc, target=target))
```
ryanhill1: **Correctness bug — this is why the converter can't be enabled as-is.** `transpile()` defaults to qiskit's `optimization_level=2`, which rewrites SWAPs into virtual qubit permutations and elides the gate, silently changing the circuit. Repro:

```python
qc = QuantumCircuit(2); qc.swap(0, 1)
qiskit_to_qrisp(qc)   # -> empty circuit; unitary no longer equal
```

`optimization_level=0` does a faithful basis translation instead. I verified swap/ccx/crx/u/parametrized/swap+gates all round-trip `unitary_equal=True` with this change:

```suggestion
    return convert_from_qiskit(transpile(qiskit_qc, target=target, optimization_level=0))
```

qbraid/transpiler/conversions/qiskit/qiskit_to_qrisp.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Args:
+        qiskit_qc: Qiskit circuit to convert to a Qrisp circuit.
+
+    Returns:
+        Qrisp Circuit equivalent to the input Qiskit circuit.
+    """
+    from qrisp.interface.converter import convert_from_qiskit  # type: ignore
+    from qiskit import transpile
+    from qiskit.transpiler import Target
```
ryanhill1: Format-check (pylint) fails here with `C0415 import-outside-toplevel`. These imports are deliberately lazy (optional deps), so silence them inline — same pattern already used at `qiskit_extras.py:83`:

```suggestion
    # pylint: disable=import-outside-toplevel
    from qrisp.interface.converter import convert_from_qiskit  # type: ignore
    from qiskit import transpile
    from qiskit.transpiler import Target
    # pylint: enable=import-outside-toplevel
```

qbraid/transpiler/conversions/qiskit/__init__.py
```diff
@@ -33,6 +33,7 @@
 from .qiskit_extras import qiskit_to_braket, qiskit_to_ionq, qiskit_to_pyqir
 from .qiskit_to_qasm2 import qiskit_to_qasm2
 from .qiskit_to_qasm3 import qiskit_to_qasm3
+# from .qiskit_to_qrisp import qiskit_to_qrisp
```
ryanhill1: To include `qiskit_to_qrisp` properly, uncomment the import:

```suggestion
from .qiskit_to_qrisp import qiskit_to_qrisp
```

and re-add it in the two spots outside this hunk:
- the autosummary block (after `qiskit_to_ionq`, ~line 30): add `   qiskit_to_qrisp`
- `__all__` (~line 43): add `    "qiskit_to_qrisp",`

With `qrisp` added to `autodoc_mock_imports` (see review summary §2), the `qrisp_` annotation in this file will resolve for the docs build. Pair this with the `optimization_level=0` fix so the enabled converter is actually correct.

tests/programs/test_program_circuits_qrisp.py
```diff
@@ -0,0 +1,87 @@
...
+    from qrisp import QuantumCircuit
+    from qbraid.programs.gate_model.qrisp import QrispCircuit
+
+    qrisp_not_installed = False
+except ImportError:
+    qrisp_not_installed = True
+
+pytestmark = pytest.mark.skipif(qrisp_not_installed, reason="qrisp not installed")
+
+from qbraid.programs.exceptions import ProgramTypeError
```
ryanhill1: Format-check (pylint) fails with `C0413 wrong-import-position` on line 32 — `ProgramTypeError` is imported after the `try`/`pytestmark` block. It doesn't depend on qrisp, so move it above the guarded imports:

```suggestion
import pytest

from qbraid.programs.exceptions import ProgramTypeError

try:
    from qrisp import QuantumCircuit

    from qbraid.programs.gate_model.qrisp import QrispCircuit

    qrisp_not_installed = False
except ImportError:
    qrisp_not_installed = True

pytestmark = pytest.mark.skipif(qrisp_not_installed, reason="qrisp not installed")
```

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,190 @@
...
+    target_circuit = transpile(cirq_circuit, target)
```
ryanhill1: This path only ever goes `qrisp→cirq` (line 160) and then `cirq→target` (line 163, the **cirq** converters) — so the direct `qrisp_to_qiskit`, `qrisp_to_pytket`, `qrisp_to_qasm2`, `qrisp_to_qasm3` (and `qiskit_to_qrisp`) functions are never exercised. That's the codecov gap: `qrisp_to_qiskit.py:46` and `qrisp_to_pytket.py:46` are uncovered (return bodies), and `qiskit_to_qrisp.py` is 0%. It's also why the missing qasm2/qasm3 registration slipped through earlier.

Suggest adding small direct round-trip tests that call each converter through `transpile(src, "<target>")` with `conversion_graph=graph`, e.g. `qrisp→qiskit/pytket/qasm2/qasm3` and a `qiskit→qrisp` case. **Include a circuit with a SWAP in the `qiskit→qrisp` test** so it guards the `optimization_level=0` fix. A quick `_unitary()` assertion would also cover `gate_model/qrisp.py:67`.

ryanhill1 (changes requested):
Thanks for the quick turnaround! I re-pulled the branch and re-ran everything in a clean **py3.11 + qrisp 0.8.2** venv. The earlier round of fixes all check out:

- ✅ `requirements-dev.txt` / `pyproject.toml` version markers, README comma, `remove_idle_qubits` index remap (+ non-trailing-idle test), `qrisp_to_qasm2/3` now registered (`qrisp` edges: `cirq, pytket, qasm2, qasm3, qiskit`), `qrisp_to_cirq` lazy import, and the `skipif` guard — all good. 7/7 qrisp tests pass.

Below are the remaining items, all tied to the three red CI checks (docs build, format-check, codecov) plus the coverage-test root cause. Details in the inline comments.

### 1. The coverage failures are a real, fixable bug — not a true conversion gap
`CPGate`/`MCRXGate`/`MCXGate` aren't unsupported: the `qrisp→cirq` conversion is unitarily exact (round-trips `allclose=True`). The `to_qasm()` crash is a **dtype interop bug** — qrisp's `to_cirq()` emits gate exponents as `numpy.longdouble`, and cirq's QASM exporter only special-cases `isinstance(value, (float, int))` (`np.float64` subclasses `float`; `np.longdouble` does **not**), so the `half_turns` format spec falls through and raises. Casting exponents to plain `float` in `qrisp_to_cirq` fixes all three (verified) and lets the baseline go back to `1.0` instead of the brittle `0.9`/`0.8929` margin. See the `qrisp_to_cirq.py` comment.

### 2. Docs build (`docs/conf.py` — not in this diff, so flagging here)
`qrisp` is missing from `autodoc_mock_imports`, while `cirq`/`qiskit`/`pytket` are all present. Because qrisp is neither installed nor mocked in the docs env, autodoc can't resolve the `qrisp_.QuantumCircuit` annotations (the `import qrisp as qrisp_` alias only exists under `TYPE_CHECKING`), which cascades into the `name 'qrisp_' is not defined` warning and the `Unknown target name: "qrisp"/"qiskit"` docutils errors. The identical `import qiskit as qiskit_` pattern in `qasm3_to_qiskit.py` builds fine *only because qiskit is mocked*. Fix — add to `docs/conf.py`:
```python
    "qiskit_ibm_runtime",
    "qrisp",
    "scc",
```

### 3. `qiskit_to_qrisp` — let's get it correctly included (don't drop it)
It works for simple circuits but has a **silent correctness bug**: `transpile(qiskit_qc, target=target)` runs at qiskit's default `optimization_level` (2), which rewrites SWAPs into virtual qubit permutations and removes the gate — `qiskit_to_qrisp(QuantumCircuit(2).swap(0,1))` returns an **empty circuit**. `optimization_level=0` preserves it; I re-tested swap/ccx/crx/u/params/swap+gates and all round-trip `unitary_equal=True`. This is almost certainly why it was commented out. The inline comments cover the fix + re-wiring + the pylint/codecov pieces.

### 4. Format-check (pylint)
- `qiskit_to_qrisp.py:46-48` → C0415 import-outside-toplevel (inline comment).
- `tests/programs/test_program_circuits_qrisp.py:32` → C0413 wrong-import-position (inline comment).

### 5. Codecov (83.83% vs 97.22%) — measured the exact uncovered lines
`qiskit_to_qrisp.py` 0% (13/13), `qrisp_to_qiskit.py:46` and `qrisp_to_pytket.py:46` (return bodies never run), `gate_model/qrisp.py:67` (`_unitary` untested). Root cause for the converter bodies: the coverage test only ever routes `qrisp→cirq` then `cirq→target`, so the direct `qrisp_to_qiskit/pytket/qasm2/qasm3` (and `qiskit_to_qrisp`) functions are never exercised. See the test comment for the suggested additions.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,185 @@
...
+    return ConversionGraph(require_native=True)
+
+
+def is_package_installed(package_name: str) -> bool:
+    """Check if a package is installed."""
+    return importlib.util.find_spec(package_name) is not None
+
+
+ALL_TARGETS = [("cirq", 1.0), ("pytket", 1.0), ("qiskit", 1.0), ("qasm2", 1.0), ("qasm3", 1.0)]
+AVAILABLE_TARGETS = [(name, version) for name, version in ALL_TARGETS if is_package_installed(name)]
```
ryanhill1: Adding `qasm2`/`qasm3` here doesn't actually test them, so `qrisp_to_qasm2.py` and `qrisp_to_qasm3.py` stay uncovered (the codecov gap):

1. **They're silently filtered out** — `is_package_installed("qasm2")` / `("qasm3")` is `False` (no such importable module), so `AVAILABLE_TARGETS` only ever contains `cirq, pytket, qiskit`. I confirmed only those three parametrizations run.
2. **Un-gating them doesn't work either** — for a string target, `convert_from_qrisp_to_x` ends with `target_circuit = transpile(cirq_circuit, target)` (a QASM *string*) and then `circuits_allclose(cirq_circuit, target_circuit)`, which can't compare a circuit to a string. I verified that forcing `qasm2`/`qasm3` through this harness scores **0%**.

The converters themselves are fine (verified `qrisp→qasm2/qasm3` produce valid QASM and round-trip `allclose=True`). To actually cover them, drop the two entries here and add a small dedicated test, e.g.:

```python
@pytest.mark.parametrize("target", ["qasm2", "qasm3"])
def test_qrisp_to_qasm_roundtrip(target, conversion_graph):
    qc = qrisp.QuantumCircuit(2)
    qc.h(0)
    qc.cx(0, 1)
    qasm = transpile(qc, target, conversion_graph=conversion_graph)
    assert isinstance(qasm, str)
    back = transpile(qasm, "cirq", conversion_graph=conversion_graph)
    assert circuits_allclose(qc.to_cirq(), back, strict_gphase=False)
```

ryanhill1 (changes requested):
## Re-review — most requested changes are in; two remain

Pulled the latest and re-verified locally (CI is stuck on `action_required` for the fork, so it hasn't actually run yet). Great progress — the substantive items are done and verified in py3.11 + qrisp 0.8.2:

- ✅ **`qrisp_to_cirq`** longdouble→float recast — coverage tests pass at **baseline 1.0**.
- ✅ **`qiskit_to_qrisp` `optimization_level=0`** + `test_qiskit_to_qrisp_with_swap` (asserts `count_ops()=={"swap":1}`) — fixes the silent SWAP-drop.
- ✅ **pylint** C0415 (disable/enable block) and C0413 (import reorder) — `pylint` 10/10 on the changed files; **black** (pinned `25.12.0`) and **isort** pass.
- ✅ **`qiskit_to_qrisp` re-enabled** (import + `__all__` + autosummary); `qrisp→qiskit/pytket` converter bodies + `_unitary()` now covered.

Two items still block green CI:

### 1. Docs build still fails — needs `qrisp` in `autodoc_mock_imports`
`docs/conf.py` isn't part of this PR's diff (the `sympy` entry came in via the `main` merge, not this branch), so I can't leave an inline comment — flagging here. With `qiskit_to_qrisp` now enabled and the `qrisp_to_*` modules autodoc'd, Sphinx tries to resolve their `qrisp_.QuantumCircuit` annotations (the `import qrisp as qrisp_` alias only exists under `TYPE_CHECKING`); since qrisp is neither installed nor mocked in the docs env, it raises `Cannot resolve forward reference ... name 'qrisp_' is not defined` plus the docutils `Unknown target name: "qrisp"/"qiskit"` errors — fatal under `sphinx-build -W`.

I reproduced the CI docs build in a clean qrisp-free env: **fails (exit 1)** as-is, and **succeeds (exit 0)** once `"qrisp"` is added to the mock list. Please add it to `docs/conf.py`:
```python
autodoc_mock_imports = [
    ...
    "qiskit",
    "qiskit_ibm_runtime",
    "qrisp",
    ...
]
```

### 2. qasm2/qasm3 coverage entries are a no-op (and can't work as written)
See the inline comment — they're silently filtered out, and even un-gated the harness scores 0% on them, so `qrisp_to_qasm2/qasm3` stay uncovered.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
ACE07-Sev: That's odd, technically this line should force it to use the direct conversion since it's the shortest path
```py
circuit = transpile(source_circuit, target, conversion_graph=graph) # force direct conversion from qrisp to target
circuit = transpile(circuit, "cirq", conversion_graph=graph) # then take target to cirq via whatever path, irrelevant
```
and did check to ensure those converters got called via print statemenets.

Nonetheless, I can add the new unit test if that's the preference.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,199 @@
...
+    """Construct a Qrisp circuit with the given gate, transpile it to
+    target program type, and check equivalence.
+    """
+    source_circuit = circuits[circuit_name]
+    circuit = transpile(source_circuit, target, conversion_graph=graph)
+    circuit = transpile(circuit, "cirq", conversion_graph=graph)
+    qasm = circuit.to_qasm()
+    cirq_circuit = circuit_from_qasm(qasm)
+    target_circuit = transpile(cirq_circuit, target)
+    assert circuits_allclose(cirq_circuit, target_circuit, strict_gphase=False)
```
ryanhill1: This helper diverges from every other `test_coverage_from_<package>.py` (braket/cirq/qiskit/pyquil/pytket), which all do a single direct conversion and compare the **original source** against the target:

```python
target_circuit = transpile(source_circuit, target, conversion_graph=graph)
assert circuits_allclose(source_circuit, target_circuit, strict_gphase=False)
```

Here, line 154's direct result is immediately overwritten on line 155, and the circuit is then routed `target → cirq → QASM → cirq → target`. The final assertion compares `cirq_circuit` against `transpile(cirq_circuit, target)` — i.e. two circuits both derived from the **same** cirq intermediate. The original qrisp `source_circuit` never enters the comparison, so this is really a cirq↔target round-trip check, not a qrisp→target fidelity check, and the QASM hop quietly normalizes away differences (dropped barriers/measurements, re-derived params).

`circuits_allclose` accepts a qrisp circuit directly (verified for cirq/qiskit/pytket), so this can match the other modules:

```suggestion
    target_circuit = transpile(source_circuit, target, conversion_graph=graph)
    assert circuits_allclose(source_circuit, target_circuit, strict_gphase=False)
```

This pairs with excluding the non-unitary gates (gate-set comment) and lowering the per-target baselines to keep RGate in (RGate comment + `ALL_TARGETS`): cirq/qiskit ≈ 0.95, pytket ≈ 0.83.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,199 @@
...
+        attr: None
+        for attr in dir(qrisp.circuit.standard_operations)
+        if attr[0] in string.ascii_uppercase
+    }
+
+    qrisp_gates.pop("Operation")
+    qrisp_gates.pop("QubitAlloc")
+    qrisp_gates.pop("QubitDealloc")
+    qrisp_gates.pop("PauliGate")
+    qrisp_gates.pop("Reset")
```
ryanhill1, line 107: Once the helper compares the qrisp source directly, the **non-unitary / special** gates need to be excluded from the benchmark — just like the other modules do (e.g. the qiskit coverage test uses `get_qiskit_gates(seed=0, exclude=["GlobalPhaseGate"])`). `circuits_allclose` can't handle them: `Measurement` raises `Unitary of operation measure not defined`, and `Barrier`/`GPhaseGate` have no meaningful unitary. They only "pass" today because the cirq+QASM detour drops them.

(Note: this is only for the genuinely non-unitary gates — **RGate stays in** and is handled via a lowered baseline, see the RGate comment.)

```suggestion
    qrisp_gates.pop("Operation")
    qrisp_gates.pop("QubitAlloc")
    qrisp_gates.pop("QubitDealloc")
    qrisp_gates.pop("PauliGate")
    qrisp_gates.pop("Reset")

    # Non-unitary / special gates aren't comparable via circuits_allclose; exclude them
    # from the benchmark, as the other coverage modules do.
    qrisp_gates.pop("Measurement")
    qrisp_gates.pop("Barrier")
    qrisp_gates.pop("GPhaseGate")
```

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,199 @@
...
+    "XXYYGate": {"qubits": [0, 1]},
+    "Barrier": {"qubits": [0]},
+    "Measurement": {"qubits": [0], "clbits": [0]},
+    "Reset": {"qubits": [0]},
+    "QubitAlloc": {"qubits": [0]},
+    "GPhaseGate": {"qubits": [0]},
+    "SwapGate": {"qubits": [0, 1]},
+    "U1Gate": {"qubits": [0], "phi": np.random.rand() * 2 * np.pi},
+    "IDGate": {"qubits": [0]},
+    "RGate": {
```
ryanhill1, line 84: RGate conversion fails the unitary equivalence check, but seems to be converted correctly — `qrisp.to_qiskit()`/`to_cirq()` both match the reference `R(θ,φ)` exactly. The wrong unitary comes from qrisp's **`get_unitary()`** (the source reference the benchmark compares against):

```
reference R(θ,φ):   [[-0.442, -0.343+0.829j], [ 0.343+0.829j, -0.442]]
qrisp .to_qiskit(): [[-0.442, -0.343+0.829j], [ 0.343+0.829j, -0.442]]   # correct ✓
qrisp get_unitary():[[-0.442, -0.829-0.343j], [ 0.829-0.343j, -0.442]]   # wrong (φ ≈ π/2 off)
```

So this might be an **upstream qrisp `get_unitary()` bug**, not a qBraid conversion problem. Keep RGate in the benchmark and lower the per-target baseline (`cirq`/`qiskit` ≈ 0.95) to account for it — and it's worth filing upstream with qrisp.

ryanhill1 (changes requested):
## Re-review — thanks for the updates, this is close!

The latest round looks great — docs build, format, the qasm round-trip coverage, the swap fix, and the longdouble recast are all sorted and verified locally. Really nice work.

One remaining thing on the coverage benchmark before this is good to go. `tests/transpiler/qrisp/test_coverage_from_qrisp.py` validates conversions differently from every other `test_coverage_from_<package>.py` module:

- The other modules do a single direct `transpile(source, target)` and compare the **original source** against the result.
- The qrisp helper instead routes `target → cirq → QASM → cirq → target` and compares the cirq intermediate against the target — so it never checks the original qrisp circuit, and it's effectively a cirq↔target round-trip rather than a qrisp→target fidelity check, which quietly normalizes away real differences.

Aligning it to the direct pattern (`circuits_allclose` accepts a qrisp circuit directly) surfaces a few things to handle:

1. **Non-unitary/special gates** (`Measurement`/`Barrier`/`GPhaseGate`) can't be unitary-compared — exclude them from the benchmark, exactly as the other modules do.
2. **RGate** fails — but the *conversion is actually correct*; the wrong unitary is in qrisp's upstream `get_unitary()` (the source reference), not in the qBraid conversion. Per your call: keep RGate in the benchmark and **lower the baseline** to acknowledge it can't be validated by unitary equivalence, rather than excluding it and claiming 1.0.

Measured per-target with the direct pattern over the unitary-only gate set (25 gates):
- **cirq: 24/25 = 0.96** — only RGate (upstream `get_unitary`)
- **qiskit: 24/25 = 0.96** — only RGate (upstream `get_unitary`)
- **pytket: 21/25 = 0.84** — RGate **plus** CPGate / U1Gate / MCRXGate

Heads up on pytket: those extras *pass* for cirq/qiskit, and `get_unitary()` is target-independent, so they're real upstream `to_pytket` discrepancies on controlled/phase gates — worth a separate look (ideally an upstream qrisp issue) rather than silently baselining away.

So suggested baselines: `cirq`/`qiskit` ≈ 0.95, `pytket` ≈ 0.83 (the `ALLOWANCE = 0.01` gives headroom). Once the benchmark is aligned, the non-unitary gates excluded, and the baselines set to reflect reality (keeping RGate in), this should be good to go. Thanks again!

ryanhill1 (commented):
_(Superseded — folded into the review above. A submitted comment-review can't be deleted, apologies for the noise.)_

ACE07-Sev:
@ryanhill1 I'll have a look at the qrisp issues and see if I can make a PR soon there.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -0,0 +1,201 @@
...
+Benchmarking tests for Qrisp conversions
+
+"""
+
+import importlib.util
+import string
+
+import numpy as np
+import pytest
+from cirq.contrib.qasm_import import circuit_from_qasm
```
ryanhill1:
```suggestion
```

ryanhill1 (approved):
Congratulations, @ACE07-Sev — this is fantastic work! 🎉 Adding Qrisp as a fully supported framework, with six new conversions wiring it into the transpiler graph, is exactly the kind of comprehensive contribution this bounty was looking for.

There were three strong PRs submitted for this issue, and after careful consideration yours was selected as the best — the breadth of a brand-new framework plus six conversions really stood out. Per the Unitary Hack rules I can only award the bounty to a single PR, and yours is the one. 🏆

Thank you for the care you put into the implementation and tests, and for your responsiveness throughout the review. I've also gone ahead and resolved the merge conflicts with `main`, so this is ready to go. Congratulations again, and thank you for such an outstanding contribution — we'd love to see more from you in the future!

ACE07-Sev:
Thank you as always Ryan.

positr0nium:
Appreciated, @ACE07-Sev

PR #1216, opened 2026-06-06
Fix Cirq → pyQuil transpilation of XX/YY/ZZ/SwapPow two-qubit gates

### LINKED ISSUE #386: Cirq --> pyQuil transpiler error for XX, YY, ZZ gates
There is a bug transpiling Cirq circuits containing the `XXPowGate`, `YYPowGate` or `ZZPowGate` to pyQuil. Below is a self-contained example showing the bug for the `XXPowGate`:

```python
import cirq

from qbraid import circuit_wrapper

# Create a quantum circuit with two qubits
q0, q1 = cirq.LineQubit.range(2)
cirq_in = cirq.Circuit()

# Add XX, YY, and ZZ gates to the circuit
exponents = [0.5, 1]
for theta in exponents:
    cirq_in.append(cirq.XXPowGate(exponent=theta).on(q0, q1))

pyquil_program = circuit_wrapper(cirq_in).transpile("pyquil")
cirq_out = circuit_wrapper(pyquil_program).transpile("cirq")

cirq_roundtrip_equal = circuits_allclose(cirq_in, cirq_out)

print(f"Cirq input:\n{cirq_in}\n")
print(f"Cirq -> pyQuil:\n{pyquil_program}\n")
print(f"pyQuil -> Cirq:\n{cirq_out}\n")
print(f"Cirq expected output:\n{cirq_in}\n")
print(f"Cirq roundtrip unitaries equal: {cirq_roundtrip_equal}")
```
```
Cirq input:
0: ───XX───────XX───
      │        │
1: ───XX^0.5───XX───

Cirq -> pyQuil:
RX(pi/2) 0
RX(pi/2) 1
X 0
X 1


pyQuil -> Cirq:
0: ───Rx(0.5π)───X───

1: ───Rx(0.5π)───X───

Cirq expected output:
0: ───XX───────XX───
      │        │
1: ───XX^0.5───XX───

Cirq roundtrip unitaries equal: False
```

ryanhill1:
## Summary

Fixes #386.

Cirq → pyQuil transpilation of the two-qubit interaction gates `XXPowGate`, `YYPowGate`, and `ZZPowGate` was incorrect for non-integer exponents. The fallback branch (inherited from `cirq_rigetti`) decomposed e.g. `XX**t` into two *independent* single-qubit rotations (`RX(πt) 0; RX(πt) 1`), which does not reproduce the two-qubit unitary. Reproduced on `main` — all three round-trip to `False`.

## Fix

`ZZPowGate(t) = diag(1, e^{iπt}, e^{iπt}, 1)` factors **exactly** (including global phase, since the (0,0) entry is a true `1`) into supported Quil primitives:

- **ZZ**: `PHASE(πt)` on each qubit + `CPHASE(-2πt)`
- **XX**: `H⊗H` · ZZ-core · `H⊗H`
- **YY**: `RX(π/2)⊗RX(π/2)` · ZZ-core · `RX(-π/2)⊗RX(-π/2)`

The integer-exponent fast paths (`X⊗X`, `Y⊗Y`, `Z⊗Z`) were already correct and are preserved.

Two constraints shaped the design:
- The pipeline is `cirq → Quil string → parse`, and the Quil input parser's `SUPPORTED_GATES` has no `RZZ`/`RXX`/`RYY`, so only `PHASE`/`CPHASE`/`H`/`RX` are usable.
- Raw `* np.pi` floats are emitted rather than the `exponent_to_pi_string` helper, which does `limit_denominator(12)` and would silently approximate arbitrary exponents (e.g. `0.37 → 3/8`).

All three gates now round-trip **exactly with `strict_gphase=True`**.

## Testing

- Added parametrized regression test `test_cirq_to_pyquil_ising_gate_roundtrip` (3 gates × 6 exponents) asserting `strict_gphase=True`.
- Updated the `test_all_operations` golden-string test to the corrected output.
- pyQuil suite: 63 passed. Combined pyQuil + cirq: 203 passed, 4 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Cirq → pyQuil transpilation for XX, YY, ZZ, and Swap power gates with non-integer exponents. Gates now correctly preserve unitary behavior during transpilation.

* **Tests**
  * Added regression tests validating Cirq-to-pyQuil round-trip conversions for the affected gates across various exponents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,7 +50,7 @@ print(result.data.get_counts())
 ### Removed
 
 ### Fixed
-- Fixed Cirq → pyQuil transpilation of the `XXPowGate`, `YYPowGate`, and `ZZPowGate` interaction gates for non-integer exponents. These were previously decomposed into independent single-qubit rotations, producing a circuit whose unitary did not match the input. They now round-trip exactly (including global phase) via `PHASE`/`CPHASE` decompositions. ([#386](https://github.com/qBraid/qBraid/issues/386))
+- Fixed Cirq → pyQuil transpilation of the `XXPowGate`, `YYPowGate`, `ZZPowGate`, and `SwapPowGate` two-qubit gates for non-integer exponents. The interaction gates were previously decomposed into independent single-qubit rotations, and `SwapPowGate` was emitted as `PSWAP` (a parametric swap-with-phase), both producing a circuit whose unitary did not match the input. The interaction gates now round-trip exactly (including global phase) via `PHASE`/`CPHASE` decompositions, and `SwapPowGate` falls back to cirq's `CNOT`/`RY`/`CPHASE` decomposition. ([#386](https://github.com/qBraid/qBraid/issues/386))
 - Fixed `qasm2_to_cirq` corrupting cirq's shared OpenQASM lexer: the QASM 2 parser assigned a reduced token list onto `cirq.contrib.qasm_import._lexer.QasmLexer.tokens` at import time, stripping the OpenQASM 3 tokens (e.g. `STDGATESINC`) process-wide and causing `qasm3_to_cirq` to raise a ply `LexError` on any QASM 3 parse that followed a `qasm2_to_cirq` call. The reduced token set now lives on a local `QasmLexer` subclass, leaving cirq's class intact ([#1214](https://github.com/qBraid/qBraid/pull/1214))
 
 ### Dependencies
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ print(result.data.get_counts())
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


ryanhill1:
Extended this PR to also cover **`SwapPowGate`** and fix the failing `format-check`.

**`SwapPowGate`** had the same class of bug as the interaction gates: for non-integer exponents it was emitted as `PSWAP(pi*t)`, but Quil's `PSWAP` is a parametric swap-with-phase, not a fractional `SWAP**t`, so the unitary didn't match. `_swappow_gate` now returns `None` for non-integer powers, falling back to cirq's `CNOT`/`RY`/`CPHASE` decomposition (the integer `SWAP` fast path is unchanged). Added `test_cirq_to_pyquil_swappow_roundtrip` (`strict_gphase=True`) and updated the `test_all_operations` golden output.

With this, `cirq → pyquil` gate coverage goes from 0.766 → **0.875**; the only remaining failures are non-unitary noise channels (`*Channel`, `ResetChannel`), which can't be validated by unitary equivalence and are outside this PR's scope.

**`format-check`** was failing because black wanted to collapse the `_xxpow_gate` implicit string concatenation onto one line, which pylint then flags as same-line `implicit-str-concat` (W1404). Resolved by making it a single string literal — satisfies both. Verified locally: `black --check`, `isort --check-only`, `ruff check`, and `pylint` (10.00/10) all pass; full pyQuil suite green (72 passed).

PR #1239, opened 2026-06-23
fix: model Rigetti device availability from QCS maintenance calendar

### LINKED ISSUE #1237: [Bug] Incorrect Rigetti Cepheus 108 availability due to bug in `status` method
### Feature description

The current availability metrics for direct integration need to be updated and get the availability of Cepheus 108 device from the Rigetti QCS API. Current the qBraid SDK implementation only checks for [whether the device is correctly being returned in the list of quantum processors](https://github.com/qBraid/qBraid/blob/main/qbraid/runtime/rigetti/device.py#L91) and implicitly assumed to be available - 

https://github.com/user-attachments/assets/7e20a5c8-6468-49c3-b286-a48cf3440cac

### Feature area

provider integration

### Motivation / use case

This will allow correct queue times to be available for the directly integrated Rigetti device 

### Alternatives considered (optional)

_No response_

### Suggested implementation (optional)

```python

```

TheGupta2012:
## Summary

Fixes the bug in `RigettiDevice.status()` reported in #1237. The method previously only checked whether the processor appeared in `list_quantum_processors()` and then always returned `ONLINE`, so a device under scheduled maintenance (e.g. Cepheus-1-108Q) was still reported as available.

`status()` now also consults Rigetti's maintenance schedule and reports:

- `OFFLINE` — processor absent from the QCS catalog.
- `UNAVAILABLE` — a maintenance window is currently in progress. During maintenance the QCS gateway queues jobs rather than executing them, so the device is reachable but not running programs (consistent with how IonQ/OQC map calibration/out-of-window states).
- `ONLINE` — otherwise.

If the calendar lookup or parsing fails, `status()` degrades to `ONLINE` (catalog membership still holds) and logs a warning, so a transient calendar-service issue never makes `status()` raise.

## Data source

`qcs_sdk` is execution-only and exposes no maintenance/availability route, so the schedule is fetched from the QCS REST API at `GET {api_url}/v1/calendars/{quantum_processor_id}`, which returns a `maintenanceICal` field — an RFC 5545 calendar of the windows during which execution is unavailable. The request reuses the device's `QCSClient` for the API base URL and the auto-refreshing OAuth bearer token.

## Availability module

Adds `qbraid/runtime/rigetti/availability.py`, modeled on `qbraid/runtime/aws/availability.py`:

- `is_in_maintenance(ical_text, now=None)` — used by `status()`.
- `next_available_time(device)` — returns `(is_available, "HH:MM:SS"_until_switch, switch_datetime)`, mirroring the AWS function. Contiguous/overlapping maintenance windows are merged so the reported "next available" time is when the device truly leaves maintenance.

`RigettiDevice` gains two methods:

- `maintenance_calendar()` — the raw maintenance iCalendar for the processor.
- `availability_window()` — delegates to `availability.next_available_time(self)` (mirrors `BraketDevice.availability_window()`).

## Dependencies

The `rigetti` extra now includes `icalendar` and `recurring-ical-events` for RFC 5545 parsing and recurrence handling.

## Testing

- New `tests/runtime/rigetti/test_availability.py` covering the helpers, `is_in_maintenance`, and `next_available_time` (online/offline/in-maintenance/contiguous-merge).
- Reworked status tests and added maintenance-calendar tests in `tests/runtime/rigetti/test_rigetti_device.py`.
- Full Rigetti suite passes (185 tests); `tox -e format-check` passes (pylint 10.00/10, isort, black, ruff, headers).
- Validated end-to-end against the real Cepheus-1-108Q maintenance windows currently published by QCS.

Closes #1237.

## Release Notes

* **New Features**
  * Rigetti devices now track maintenance schedules with real-time availability status
  * Device status now displays `UNAVAILABLE` during scheduled maintenance, `OFFLINE` when unavailable, and `ONLINE` otherwise
  * Added maintenance calendar and next-available-time information for Rigetti devices

* **Bug Fixes**
  * Maintenance calendar lookup and parsing failures now gracefully degrade to `ONLINE` status with a warning

* **Dependencies**
  * Added `icalendar` and `recurring-ical-events` for RFC 5545 maintenance calendar support


ryanhill1 (changes requested):
Bring patch coverage up >=97.20% and then this is good to go

ryanhill1 (approved):

PR #1244, opened 2026-06-27
fix: cached_method unhashable args and instance-scoped cache keys; skip unbuildable Braket device profiles

ryanhill1:
## Summary

Three coupled fixes: two `cached_method` bugs and one `BraketProvider.get_devices` bug. Together they (a) let `get_devices` include retired devices via `statuses=["ONLINE", "OFFLINE", "RETIRED"]` (currently impossible — see below), and (b) stop long-lived processes from serving permanently stale device data (root cause of the qbraid-runtime-api device-status sync writing outdated statuses, e.g. AQT Ibex-Q1 stuck ONLINE while OFFLINE in the Braket console).

### 1. `cached_method` rejects unhashable arguments

`cached_method` routed every call through an inner `functools.lru_cache`, which re-hashes the raw arguments and raises `TypeError: unhashable type: 'list'` when a decorated method is called with a `list`/`dict`:

```python
provider.get_devices(statuses=["ONLINE", "OFFLINE", "RETIRED"])  # TypeError
```

The wrapper already computes a JSON-derived cache key via `_generate_cache_key` and maintains its own TTL dict, so the inner `lru_cache` was redundant — and the only thing requiring hashable args. The decorator now caches solely against the JSON-derived key, so unhashable arguments are supported.

Removing the inner `lru_cache` also fixes a **permanent-staleness bug**: on TTL expiry the wrapper "recomputed" by calling the lru-cached inner function, whose key is the provider instance. Since `QuantumProvider.__eq__` treats all same-class providers as equal and providers hash their credentials, every fresh provider hit the same lru entry — so the first result computed in a process's lifetime was returned forever, merely re-stamped as fresh. TTL expiry now calls the real function.

Preserved: TTL expiry, `maxsize` bounding (oldest-by-timestamp eviction), `cache_clear()`, `cache_info().currsize`, the `DISABLE_CACHE` env switch, and the `cache_disabled` context manager. Removed: the `_QBRAID_TEST_CACHE_CALLS` test hook, which existed only to work around `lru_cache` hiding `func` calls from mock call-counts — unnecessary now that every cache miss is a direct `func()` call.

### 2. `cached_method` cache keys omit instance identity

`_generate_cache_key` hashed only the class name (plus method/args), so two instances of the same provider class configured with **different credentials** shared cache entries within the TTL — one provider could be served another's cached results. The key now includes `hash(instance)`: providers hash their credentials, so equal credentials still share the cache, while different credentials get separate entries. Classes that are unhashable (`__eq__` without `__hash__`, like the `QuantumProvider` base) fall back to `id(instance)`, degrading to per-instance caching rather than raising.

### 3. `get_devices` aborts on a single unbuildable profile

`BraketProvider.get_devices` builds a runtime profile for **every** returned device in one comprehension. A device without a qBraid-supported program type (no `braket.ir.openqasm.program` / `braket.ir.ahs.program`) makes `_build_runtime_profile` raise `QbraidError`, aborting the whole call. This is reachable today with `statuses=["...","RETIRED"]`, which surfaces legacy hardware such as D-Wave `Advantage_system1`. Such devices are now skipped (logged at info level) and the remaining devices are returned.

## Why

Downstream (qbraid-runtime-api) syncs Amazon Braket device statuses via `BraketProvider.get_devices()`, which defaults to `["ONLINE", "OFFLINE"]`. Two production symptoms trace here:

- RETIRED devices (e.g. TN1) silently drop out of the sync and freeze at their last status. Passing `statuses=[..., "RETIRED"]` is the intended fix, but issues 1 and 3 block it.
- Device statuses freeze at whatever the Cloud Run worker first fetched (issue 1's staleness bug): the prod sync ran every 5 minutes yet kept writing ONLINE for AQT Ibex-Q1 days after it went OFFLINE, with a fresh `updatedAt` each run (queue depth stayed correct because `queue_depth()` makes a live API call, while `.status` reads the frozen cached device object).

## Tests

- `test_caching.py`: new `test_cached_method_accepts_unhashable_args` (list arg cached, distinct lists are distinct keys, `cache_info`/`cache_clear`), `test_cache_key_separates_instances_with_different_hashes` (different-credential instances don't share entries; equal-hash instances do), and `test_cache_key_falls_back_to_id_for_unhashable_instances`. Existing `test_clear_cache` still green.
- `test_braket_runtime.py`: new `test_provider_get_devices_skips_unbuildable_profiles` (a device that raises in `_build_runtime_profile` is skipped, the rest returned).
- `test_device.py`: dropped the now-defunct `_QBRAID_TEST_CACHE_CALLS` setenv from two cache tests; they assert the real call-counts unchanged.

Verified against the live Braket API: `get_devices(statuses=["ONLINE","OFFLINE","RETIRED"])` now returns 21 devices including TN1 as `RETIRED`, with incompatible devices skipped. Staleness fix verified by simulating TTL expiry across fresh provider instances: pre-fix the first device list is returned forever; post-fix expiry triggers a real re-fetch. `black`/`isort`/`ruff` clean; `pylint` 10/10 on changed source; full caching + AWS runtime + device suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved caching so methods now support unhashable arguments and correctly separate results between instances.
  * Added bounded cache eviction and reliable cache expiration behavior.
  * Ensured cache-disabled mode bypasses cache reads and writes.
  * Updated device discovery to skip unsupported devices instead of failing the entire request.


ryanhill1:
@coderabbitai  review

PR #1258, opened 2026-07-12
fix(ibm): raise typed API errors so a 404 isn't mistaken for a credentials failure

ryanhill1:
## Problem

`QiskitRuntimeProvider._ibm_api_get` and `._exchange_api_key` collapse **every** failure into a bare `ValueError`, discarding IBM's HTTP status:

```python
except (URLError, OSError) as e:
    raise ValueError(f"IBM API request failed: {e}") from e
```

`HTTPError` is a subclass of `URLError` and carries `.code`, so the status is right there — and thrown away. The only way a caller can tell *why* the call failed is to parse the message string.

### How this bites users

In qBraid Lab, searching the IBM jobs tab for a **job ID that doesn't exist** produces a 404. Because the Lab can't distinguish that from an auth failure, it showed users this:

> ⚠️ IBM API request failed: HTTP Error 404: Not Found
> _Save your IBM account credentials **saved to disk** in: `~/.qiskit/qiskit-ibm.json`_

…to users whose credentials were **perfectly valid and working**. The downstream fix (qBraid/qbraid-lab-extensions#131) currently has to classify by matching substrings like `"404"`, `"not found"`, `"unauthorized"` against the exception text — which is brittle, and silently regresses if these messages are ever reworded. That's a symptom; the cause belongs here.

## Change

Typed exceptions that carry the status code, so callers branch on **type**, not prose:

| Exception | When | Notes |
|---|---|---|
| `RuntimeAPIError(message, status_code=None)` | any provider API error | exposes `.status_code` |
| `AuthorizationError` | **401 / 403** — credentials missing, expired, or rejected | subclasses `RuntimeAPIError` |
| `JobNotFoundError` | **404** — job/task doesn't exist | subclasses `RuntimeAPIError` **and** the existing `ResourceNotFoundError` |

Wired into both `_ibm_api_get` (404 / 401 / 403 / other) and `_exchange_api_key` (IAM rejecting a bad or expired API key → `AuthorizationError`, which is the real "credentials rejected" path). All three are exported from `qbraid.runtime`.

## Backwards compatibility

These paths previously raised `ValueError`, and there's an existing test asserting exactly that. To avoid breaking anyone:

- `RuntimeAPIError` subclasses **`ValueError`** → existing `except ValueError` keeps working.
- `JobNotFoundError` subclasses **`ResourceNotFoundError`** → existing handlers for that keep working.

Both bases are additive. The `ValueError` base is transitional and can be dropped in a future major — flagged in the CHANGELOG. **This is the main thing I'd like a second opinion on** (see below).

```python
JobNotFoundError("...", status_code=404)
#  isinstance -> RuntimeAPIError ✅  ResourceNotFoundError ✅  ValueError ✅  .status_code == 404
```

## Tests

7 new cases in `TestApiErrorTyping`:
- status → type mapping (404→`JobNotFoundError`, 401/403→`AuthorizationError`, 500→`RuntimeAPIError`), each asserting `.status_code`
- **a 404 is NOT an `AuthorizationError`** — the precise bug
- the `ValueError` compat path still holds
- IAM rejecting a bad API key raises `AuthorizationError`

```
tests/runtime/ibm/  67 passed
tests/runtime/     592 passed, 254 skipped — no regressions
```
(`tests/runtime/test_device.py` + `test_qir_simulator_remote.py` fail to collect on `main` too — missing optional `cirq` — unrelated.)

## Open question for reviewers

Is the `ValueError` base on `RuntimeAPIError` worth keeping? It's purely for compat and is a little unusual in a typed hierarchy. The alternatives:

1. **Keep it** (this PR) — zero breakage, ships in a patch/minor.
2. **Drop it** — cleaner hierarchy, but it's a breaking change for anyone catching `ValueError`; should then land in a minor with a note, not a patch.

Happy to switch to (2) if you'd rather take the break now.

## Summary by CodeRabbit

- **Bug Fixes**
  - IBM Quantum API failures are now reported with specific error types for authorization issues, missing jobs, and other runtime failures.
  - API errors include the relevant HTTP status code when available.
  - Existing error-handling behavior remains compatible with prior `ValueError` and not-found handling.


ryanhill1:
## Cross-checked against IBM's OpenAPI spec + live API

Validated this branch against [`quantum.cloud.ibm.com/api/openapi.json`](https://quantum.cloud.ibm.com/api/openapi.json) (Qiskit Runtime API 0.47.5) and against live responses from the real endpoint.

### The existing mapping holds up

- **Base URL is fine.** We call `us-east.quantum-computing.cloud.ibm.com/jobs` while the spec documents `quantum.cloud.ibm.com/api/v1/jobs`. Same endpoint — the legacy host rewrites internally, which it leaks in its own error text (`GET /v1/jobs` there returns ``Cannot GET /api/v1/v1/jobs``). So the spec does govern our calls.
- **Auth matches.** `Authorization: Bearer <IAM token>` + `Service-CRN` is the spec's `IBMCloudAuth` + `ServiceCRN`.
- **`get_job` mapping is complete.** `GET /v1/jobs/{id}` documents exactly 200/401/403/404, and 404 → `JobNotFoundError`, 401/403 → `AuthorizationError` is confirmed live:
  ```
  GET /jobs/deadbeef-...  ->  404  {"errors":[{"code":1291,"message":"Job not found. Job ID: ..."}]}
  ```

### Two things the spec surfaced, now fixed in 7b015e5

**1. We were throwing away a better signal than the HTTP status.** Every IBM error carries an `ErrorContainer` body, and `HTTPError` is file-like so it was there for the taking:

```json
{"errors": [{"code": 1291, "message": "Job not found. Job ID: ...",
             "solution": "Verify the job ID is correct...",
             "more_info": "https://cloud.ibm.com/apidocs/quantum-computing#error-handling"}],
 "trace": "c4dd86aa-1150-485e-85f3-2e4e5f410317"}
```

IBM's `code` is finer-grained than the status (1291 job-not-found vs 1200 not-authorized vs 1219 auth-error), and `trace` is the request id support asks for. Both now ride on the exception as `.error_code` / `.trace`.

Worth noting for the Lab side: IBM's own `solution` field is essentially the credentials guidance we were hardcoding — it can be surfaced verbatim, at the times IBM actually means it.

**2. Not every 403 comes from IBM — and this branch would have mislabeled those.** An edge proxy in front of the API answers **403 with an HTML block page**. I hit it by accident: a request with no `User-Agent` header gets blocked before it ever reaches the Runtime API, and *every* call returns 403 HTML — including ones with entirely valid credentials.

Under the mapping as it stood, that became `AuthorizationError` → "your credentials were rejected, go re-save them." Which is precisely the bug this branch exists to kill, walking back in through a different door. Now:

- 401/403 **with** an `ErrorContainer` body → `AuthorizationError` (IBM really did reject us)
- 401/403 **with a foreign body** (HTML, malformed JSON) → plain `RuntimeAPIError`
- 401/403 **with no body at all** → falls back to the status mapping, so existing behavior and tests are unchanged

**Spec drift, so the body is parsed defensively:** `ErrorContainer.code` is declared `type: string` but the API sends an **int**, and live responses include a `solution` field the schema doesn't declare.

### Things I checked that turned out fine

- I suspected a 404 on the **list** endpoint (which the spec documents, and which `_ibm_api_get` would have typed as `JobNotFoundError` despite there being no job id in the request) would fire on a bad instance CRN. It doesn't — a bogus CRN returns **403** `code: 1200`, not 404. I could not produce a 404 from `GET /jobs` at all. Leaving the mapping as-is.
- The spec documents **400** on list_jobs, but a bad filter (`limit=abc`) returns **200** — IBM ignores it. No handling needed.
- No `429` appears anywhere in the spec; the `500`s it documents fall into the `RuntimeAPIError` catch-all correctly.

### Out of scope for this spec

`_exchange_api_key` hits `iam.cloud.ibm.com/identity/token`, which is IBM Cloud IAM and appears nowhere in the Runtime spec — so the 400/401/403 → `AuthorizationError` mapping there is unverified by this exercise. Flagging that folding **400** into `AuthorizationError` means a malformed IAM request (missing `grant_type`) would also be reported to the user as a credentials failure.

Separately: the spec's `IBMCloudAPIKey` scheme accepts `Authorization: apikey <key>` directly against the Runtime API, which suggests the IAM token exchange may be avoidable altogether — that would delete this failure path rather than type it. Not for this PR.

### Tests

5 new cases in `TestIbmErrorBody`, written against the live captured payloads: error code + trace propagation, a 403 with an IBM body **is** an auth error, a 403 from an edge proxy **is not**, a malformed body is not, and a body-less error still falls back to the status mapping.

```
tests/runtime/ibm/   70 passed
tests/runtime/      639 passed, 254 skipped  (no new failures)
```

PR #1262, opened 2026-07-13
feat(runtime): add AQT provider

### LINKED ISSUE #1260: [FEATURE] AQT direct integration
## Feature description
Implement AQT direct integration in qBraid using the AQTProvider runtime.

## Feature area
- provider integration

## Motivation / use case
Adding native support for AQT as a provider will enable users to submit quantum circuits directly to AQT's quantum systems through qBraid, improving accessibility and workflow integration with AQT's quantum infrastructure.

## References
- https://arnica.aqt.eu/api/v1/docs#/
- https://www.aqt.eu/quantum-sdk-connectors/
- https://qiskit-community.github.io/qiskit-aqt-provider/

TheGupta2012:
**[New Feature]** — New Feature

Adds a native AQT (Alpine Quantum Technologies) runtime provider so qBraid users can submit to AQT arnica-cloud devices through the standard runtime interface.

## Design

The integration mirrors how the IonQ device is wired, using a `serialize` hook rather than a new transpiler program type:

- The device profile targets the existing `qiskit` alias with a serialize hook: `ProgramSpec(qiskit.QuantumCircuit, alias="qiskit", serialize=qiskit_to_aqt)`. Any qBraid-supported program is transpiled to a qiskit circuit, then the hook (`converter.py`) delegates to `qiskit-aqt-provider` (offline `translation_method="aqt"` / `scheduling_method="aqt"` for API-valid `{rz, r, rxx}` angles) to build the AQT `quantum_circuit` payload. No new transpiler program type is registered.
- **Auth** is handled by `aqt-connector` (OIDC/Auth0), resolved non-interactively from `AQT_ACCESS_TOKEN`, or `AQT_CLIENT_ID`/`AQT_CLIENT_SECRET` client credentials — the interactive device flow is never triggered.
- **Job I/O** goes through an `AQTSession` (`qbraid_core.sessions.Session`) hitting the arnica v1 REST API directly (`/workspaces`, `/resources/{id}`, `/submit/{ws}/{res}`, `/result/{job_id}`, `DELETE /jobs/{job_id}`).
- Devices are addressed by a `"<workspace>/<resource>"` `device_id`.

## Changes

- **Added**: `qbraid/runtime/aqt/` package with `AQTProvider`, `AQTDevice`, `AQTJob`, `AQTSession`, and the `qiskit_to_aqt` serialize hook (`converter.py`).
- **Registered**: `aqt` in the `qbraid.runtime` lazy-load dict and `TYPE_CHECKING` imports so the provider is discoverable from `qbraid.runtime`.
- **Packaged**: the `aqt` optional-dependency extra (`qiskit-aqt-provider>=1.15.0`, `aqt-connector`) plus `aqt` entries under the `qbraid.providers` and `qbraid.jobs` entry-point groups in `pyproject.toml`.
- **Tested**: added `tests/runtime/aqt/test_aqt_runtime.py` and registered `"aqt"` in the expected providers list in `tests/runtime/test_loader.py`.

## Testing

- `tox -e format-check` passes (pylint 10.00/10, isort, black, ruff, Apache headers).
- `tox -e unit-tests` passes (2184 passed, 428 skipped). The AQT vendor-conversion tests `importorskip` `qiskit-aqt-provider` / `aqt-connector` (kept out of `requirements-dev.txt` because adding it triggers a pip `resolution-too-deep` on Python 3.13); the full AQT suite (31 tests) passes locally in an env with the `[aqt]` extra installed.

## Related PRs

- qBraid/qbraid-runtime-api#214 — runtime-api: AQT provider dispatch (consumer)
- qBraid/qbraid-infrastructure#586 — infra: AQT env vars on the staging Cloud Run


qbraid/runtime/aqt/converter.py
ryanhill1, line 1: AQT actually defiens a circuit type: `aqt_connector.models.circuits.Circuit`. I didn't know that.  It is the return type of `qiskit_to_aqt_circuit`. So we can actually move this conversion to the transpiler since there is no dict type conflict.

qbraid/runtime/aqt/converter.py
```diff
@@ -0,0 +1,86 @@
...
+            circuit,
+            basis_gates=AQT_BASIS_GATES,
+            translation_method="aqt",
+            scheduling_method="aqt",
+            optimization_level=1,
+        )
+    return {
+        "quantum_circuit": qiskit_to_aqt_circuit(transpiled).model_dump(),
+        "number_of_qubits": circuit.num_qubits,
+    }
```
ryanhill1: This is handled in [circuits_to_aqt_job](https://github.com/qiskit-community/qiskit-aqt-provider/blob/561551713bfcdb33407eccec8b035fa88f446ef2/qiskit_aqt_provider/circuit_to_aqt.py#L131). Return type here should just be a `aqt_connector.models.circuits.Circuit` and then the circuits_to_aqt_job step (or our equivalent) can be done inside submit

qbraid/runtime/aqt/job.py
```diff
@@ -0,0 +1,139 @@
...
+    def cancel(self) -> None:
+        """Cancel the AQT job."""
+        self.session.cancel_job(self.id)
+
+    def _device_id(self, job_metadata: dict[str, Any]) -> str:
+        """Resolve the device id for the result, from the device or the job metadata."""
+        if self._device is not None:
+            return self._device.id
+        workspace_id = job_metadata.get("workspace_id", "")
+        resource_id = job_metadata.get("resource_id", "")
```
ryanhill1: are workspace_id and resource_id not guaranteed in job_metadata?

qbraid/runtime/aqt/job.py
```diff
@@ -0,0 +1,139 @@
...
+        if status != "finished":
+            message = response.get("message", "")
+            raise AQTJobError(
+                f"Job {self.id} did not finish successfully (status={status}). {message}".strip()
+            )
+
+        measurement_counts = _samples_to_counts(response.get("result", {}))
+        data = GateModelResultData(measurement_counts=measurement_counts)
+        return Result(
+            device_id=self._device_id(payload.get("job", {})),
```
ryanhill1: so this device id fallback is `"/"` ?

is it possible that the payload does not contain the "job" field? is this fallback even necessary?

qbraid/runtime/aqt/job.py
```diff
@@ -0,0 +1,139 @@
...
+        response = payload.get("response", {})
+        status = response.get("status")
+
+        if status != "finished":
+            message = response.get("message", "")
+            raise AQTJobError(
+                f"Job {self.id} did not finish successfully (status={status}). {message}".strip()
+            )
+
+        measurement_counts = _samples_to_counts(response.get("result", {}))
```
ryanhill1: so measurement counts can be empty dict with job that has success true?

qbraid/runtime/aqt/device.py
```diff
@@ -0,0 +1,132 @@
...
+        """Return the AQT session."""
+        return self._session
+
+    def __str__(self):
+        return f"{self.__class__.__name__}('{self.id}')"
+
+    @property
+    def workspace_id(self) -> str:
+        """Return the arnica workspace id for this device."""
+        return self.profile.get("aqt_workspace_id") or self.id.split("/", 1)[0]
```
ryanhill1: in the _build_profile function is looks like workspace id is required, so it should be there

qbraid/runtime/aqt/device.py
```diff
@@ -0,0 +1,132 @@
...
+
+    @property
+    def resource_id(self) -> str:
+        """Return the arnica resource id for this device."""
+        return self.profile.get("aqt_resource_id") or self.id.split("/", 1)[-1]
```
ryanhill1: in the build profile function it looks like resource_id is required, so it should be there

qbraid/runtime/aqt/device.py
```diff
@@ -0,0 +1,132 @@
...
+        except KeyError as err:
+            raise ValueError(f"Unrecognized device status: {status}") from err
+
+    # pylint: disable-next=arguments-differ
+    def submit(
+        self,
+        run_input: Union[AQTCircuitPayload, list[AQTCircuitPayload]],
+        shots: int = 100,
+        name: Optional[str] = None,
+        runtime_options: Optional[dict[str, Any]] = None,
```
ryanhill1: unused, remove

qbraid/runtime/aqt/provider.py
```diff
@@ -0,0 +1,248 @@
...
+        resource_type = resource.get("type", "")
+        return TargetProfile(
+            device_id=f"{workspace_id}/{resource_id}",
+            simulator=resource_type == "simulator",
+            experiment_type=ExperimentType.GATE_MODEL,
+            num_qubits=resource.get("available_qubits"),
+            # Target qiskit: the transpiler routes any supported program to a qiskit circuit,
+            # then the serialize hook converts it to the AQT native-basis payload.
+            program_spec=ProgramSpec(
+                qiskit.QuantumCircuit, alias="qiskit", serialize=qiskit_to_aqt
```
ryanhill1: target can be the aqt_connect Circuit now that we know that exists. serialize step can convert it to dict using model_dump

pyproject.toml
```diff
@@ -42,6 +42,7 @@ Documentation = "https://docs.qbraid.com/v2/sdk"
 Discord = "https://discord.gg/TPBU2sa8Et"
 
 [project.optional-dependencies]
+aqt = ["qiskit-aqt-provider>=1.15.0", "aqt-connector"]
```
ryanhill1: It seems like the only function we are using from the qiskit-aqt provider is [qiskit_to_aqt_circuit](https://github.com/qiskit-community/qiskit-aqt-provider/blob/master/qiskit_aqt_provider/circuit_to_aqt.py#L31). In this case, we can just copy that function into the transpiler and site it, since it is licensed under Apache 2.0, and drop this dependency. 

If we want to include the qiskit-aqt-provider dependency, I suggest we get rid of the AQTSession, and all of the other data models, and results processing, since all of that is already present in the provider class.

So in summary, either we get rid of the qiskit-aqt-provider dependency and move forward with this API spec based implementation, or we fully leverage the qiskit-aqt-provider and all of its functionality, and strip everything down to just wrap that. We are pretty close to the former. So that will probably be easiest right now.

qbraid/runtime/aqt/provider.py
```diff
@@ -0,0 +1,248 @@
...
+            ),
+            provider_name="AQT",
+            # Extras (accessible via ``device.profile.<key>``): arnica routing + metadata.
+            aqt_workspace_id=workspace_id,
+            aqt_resource_id=resource_id,
+            aqt_resource_type=resource_type,
+        )
+
+    @cached_method
+    def get_devices(self, **kwargs) -> list[AQTDevice]:  # pylint: disable=unused-argument
```
ryanhill1: remove kwargs if unused

ryanhill1 (changes requested):
- I think there are probably models in `aqt_connector.models` that we are recreating right now ourselves e.g. submit job request. probably others. Let's look at aqt_connector more closely to make sure we are utilizing their data types and not duplicating with our own equivalent ones.

- Replace all Optional and Union types with `|`

- Revisit each instance of `.get("something", "")` to make sure the fallback is really necessary. there is a lot of this and I think it is probably overly conservative

- Let's use the `aqt_connector` circuit type, move the qiskit to aqt conversion to the transpiler, and drop the qiskit-aqt-provider dependency cause it doesn't seem like were really utilizing it at all, outside of the one conversion function that we can copy over under apache 2.0

- Add aqt provider to `docs/api/qbraid.runtime.rst`

qbraid/runtime/aqt/job.py
```diff
@@ -0,0 +1,139 @@
...
+    """Convert AQT per-shot measurement samples to bitstring counts.
+
+    The arnica result maps each circuit index to a list of shots, where each shot is a list of
+    per-qubit measurement outcomes, e.g. ``{"0": [[1, 0], [1, 1], ...]}``.
+    """
+    per_circuit: list[MeasCount] = []
+    for index in sorted(result, key=int):
+        counts: MeasCount = {}
+        for sample in result[index]:
+            bitstring = "".join(str(bit) for bit in sample)
```
ryanhill1: Bit ordering here is worth double-checking against qBraid's convention. This joins qubits q0-first (q0 becomes the leftmost character), but the repo isn't uniform: `aws/result_builder.py` reverses per-qubit order ("Convert to qBraid format with reversed bitstring order") while `rigetti/job.py` doesn't. Whichever convention AQT's arnica samples follow, the current tests (`test_job_result_success`, `test_samples_to_counts_single`) assert the ordering without validating it against a known asymmetric circuit — so a reversal would pass silently. Could you confirm against e.g. an `X` on q0 only (expected `"01"` vs `"10"`) and pin it with a test?

qbraid/runtime/aqt/provider.py
```diff
@@ -0,0 +1,248 @@
...
+
+    def get_workspaces(self) -> list[dict[str, Any]]:
+        """List the workspaces (and their resources) visible to the token."""
+        return self.get("/workspaces").json()
+
+    def get_resource(self, resource_id: str) -> dict[str, Any]:
+        """Return the details (status, available qubits, characterization) of a resource."""
+        try:
+            return self.get(f"/resources/{resource_id}").json()
+        except Exception as err:  # pylint: disable=broad-except
```
ryanhill1: `except Exception` reports every failure — a 401 (bad/expired token), a 500, or a timeout — as `ResourceNotFoundError("Resource X not found")`. That's misleading to debug, and in the runtime-api an auth/network error would surface as a not-found. Consider narrowing this to an actual 404 (inspect the response status) and letting other errors propagate. `test_session_get_resource_not_found` feeds a generic `RuntimeError`, confirming everything is currently swallowed.

qbraid/runtime/aqt/converter.py
```diff
@@ -0,0 +1,86 @@
...
+        transpiled = qiskit.transpile(
+            circuit,
+            basis_gates=AQT_BASIS_GATES,
+            translation_method="aqt",
+            scheduling_method="aqt",
+            optimization_level=1,
+        )
+    return {
+        "quantum_circuit": qiskit_to_aqt_circuit(transpiled).model_dump(),
+        "number_of_qubits": circuit.num_qubits,
```
ryanhill1: `number_of_qubits` is read from the pre-transpile `circuit`, while `quantum_circuit` above comes from `transpiled`. Harmless today, but `transpiled.num_qubits` would be more robust if transpilation ever changes the register size (ancillas/layout).

qbraid/runtime/aqt/converter.py
```diff
@@ -0,0 +1,86 @@
...
+            — the arnica per-circuit payload without ``repetitions`` (added at submit time). The
+            ``number_of_qubits`` is taken directly from the circuit's ``num_qubits`` property.
+    """
+    # pylint: disable-next=import-outside-toplevel
+    from qiskit_aqt_provider.circuit_to_aqt import qiskit_to_aqt_circuit
+
+    with warnings.catch_warnings():
+        # The AQT scheduling stage warns about missing instruction durations when run without a
+        # backend; it is only used here for its angle-wrapping passes, so the warning is benign.
+        warnings.simplefilter("ignore", UserWarning)
```
ryanhill1: `simplefilter("ignore", UserWarning)` suppresses *all* UserWarnings for the duration of the transpile, not just the AQT scheduling one this is meant to hide. Consider narrowing with `warnings.filterwarnings("ignore", message=...)` so unrelated warnings still surface.

qbraid/runtime/aqt/converter.py
TheGupta2012: I did implement it initially but thought that just adding a `serialize` step was simpler. However, I agree adding to the transpiler opens the path to adding more direct conversions into `aqt` node

qbraid/programs/_import.py
```diff
@@ -155,6 +155,13 @@ def _get_class(module: str):
         "qat.core.wrappers.circuit",
     ]
 )
+try:
+    # aqt_connector's native circuit is the "aqt" transpiler node (target of the qiskit -> aqt
+    # edge). Its module name would auto-alias to "aqt_connector", so register it explicitly under
+    # "aqt". Guarded: aqt-connector is an optional extra.
+    dynamic_non_native["aqt"] = import_module("aqt_connector.models.circuits").QuantumCircuit
```
TheGupta2012: @ryanhill1 could you explain how we can register this type correctly? I'm particularly unsure if this registration is the right way to do it

TheGupta2012:
### AQT rollout — follow-up issues

Tracking the remaining AQT integration work outside this PR:
- **qbraid-account** (update provider info + seed device docs in staging): qBraid/qbraid-account#578
- **qbraid-infrastructure** (create provider secrets in staging & prod): qBraid/qbraid-infrastructure#618
- **docs** (provider onboarding documentation): qBraid/docs#284

ryanhill1 (approved):

PR #1263, opened 2026-07-13
fix(ibm): surface IAM and Quantum API error codes in HTTP failures

ryanhill1:
## Summary

`QiskitRuntimeProvider._exchange_api_key` and `_ibm_api_get` wrap `urllib` failures in `ValueError`, but formatted `HTTPError` as just its status line — discarding the response body, which is where IBM puts the actual reason:

```
ValueError: Failed to exchange IBM API key: HTTP Error 400: Bad Request
```

A new `_format_http_error` helper reads the error body and extracts both shapes IBM uses:

- **IBM Cloud IAM** (`iam.cloud.ibm.com/identity/token`): `errorCode` / `errorMessage` / `errorDetails`
- **IBM Quantum Runtime API** (per its OpenAPI `ErrorContainer` schema): `errors[].code` / `message` / `more_info`, plus the `trace` request id (useful when escalating to IBM support)

Non-JSON bodies (e.g. HTML error pages) are appended raw, truncated to 500 chars; empty/unreadable bodies degrade to the previous `str(error)` behavior. `URLError`/`OSError` handling is unchanged.

The same failure now reads:

```
ValueError: Failed to exchange IBM API key: HTTP Error 400: Bad Request: BXNIM0415E - Provided API key could not be found.
```

## Why

Multiple users hit `Failed to exchange IBM API key: HTTP Error 400: Bad Request` with no way to tell what was wrong. Debugging one such report required re-issuing the IAM request manually to recover the discarded body — which immediately identified the cause (`BXNIM0415E`: the saved API key was 43 chars instead of 44, i.e. mis-copied). IAM distinguishes invalid keys (`BXNIM0415E`), empty keys (`BXNIM0109E`), suspended accounts, etc. — all previously flattened into "Bad Request".

## Tests

- `test_surfaces_iam_error_code_on_http_error` — a 400 with an IAM body raises `ValueError` containing `BXNIM0415E` and the message (regression for the report above).
- `test_surfaces_runtime_api_error_on_http_error` — a 404 with an `ErrorContainer` body surfaces `errors[].code`/`message` and `trace`.
- `TestFormatHttpError` — fallbacks: non-JSON body (truncated), empty body, unreadable body (`fp=None`), unrecognized JSON shape.

Verified against live IBM IAM with a deliberately invalid key: the raised error now includes `BXNIM0415E - Provided API key could not be found.` Full `tests/runtime/ibm` suite passes; `tox -e format-check` clean (pylint 10/10, mypy, ruff, isort, black).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for IBM API failures.
  * Error details now include relevant error codes, messages, request IDs, and additional guidance when available.
  * Added a fallback that displays response content when structured error details cannot be identified.


ryanhill1:
Merged `main` (now including #1258's typed IBM API errors) and reworked this PR onto the typed hierarchy in `7176d68d`. The scope is narrower now:

- **Dropped**: the Runtime-API (`ErrorContainer`) message formatting — #1258's `_parse_ibm_error` already surfaces `errors[].code`/`message`/`more_info`/`trace` as structured fields on the typed exceptions.
- **Kept (this PR's unique value)**: IAM error-body parsing on the key-exchange path. `_exchange_api_key`'s `AuthorizationError`/`RuntimeAPIError` now include IAM's `errorCode`/`errorMessage`/`errorDetails` in the message (e.g. `BXNIM0415E - Provided API key could not be found.`), with `error_code` and IAM's `context.requestId` populated on the exception's `error_code`/`trace` fields. Non-JSON bodies are surfaced raw, truncated.
- **Kept**: the read-hardening — a closed/consumed body stream raises `ValueError` from `read()`, which previously escaped the parser; both `_parse_ibm_error` and the new `_parse_iam_error` now degrade to status-only mapping instead.

Tests reconciled accordingly: the IAM regression test now asserts the typed `AuthorizationError` with `error_code`/`trace`; the `_format_http_error` fallback tests became `TestParseIamError`; the redundant Runtime-API formatting test is gone (covered by #1258's `TestIbmErrorBody`). `tests/runtime/ibm/test_qiskit_jobs.py`: **50 passed** locally; black/isort clean.

PR #1270, opened 2026-07-16
fix(transpiler): raise instead of silently dropping delay in qasm3_to_braket

### LINKED ISSUE #1269: [BUG] delay instructions silently dropped when transpiling to Braket (T1/DD experiments return meaningless data)
### Summary

`delay` instructions are **silently discarded** when a circuit is transpiled to a Braket target. No exception, no warning, no log line. Circuits that depend on idle time (T1/T2 characterization, dynamical decoupling) run as zero-delay circuits and return plausible-looking but meaningless data.

This came in as a customer report: they were sweeping delay to measure T1 on `Rigetti Cepheus-1-108Q` via AWS and got a completely flat curve. The same code works on IBM. Their delays never reached the QPU — every point in the sweep was the same zero-delay circuit, and they burned QPU time before noticing.

### Environment

```
qbraid            0.12.2
amazon-braket-sdk 1.91.0
```

Affected area: transpiler / conversion. Not Lab-specific.

### Steps to reproduce

```python
from braket.circuits import Circuit

qasm = """OPENQASM 3.0;
bit[1] b;
qubit[1] q;
h q[0];
delay[100000ns] q[0];
b[0] = measure q[0];
"""

c = Circuit.from_ir(source=qasm)
print(len(c.instructions))          # 2  -- H and Measure. The delay is gone.
print(c.to_ir("OPENQASM").source)   # delay[100000ns] absent from the payload
```

Via the normal user-facing path (`qiskit.QuantumCircuit.delay(...)` → `transpile(..., "braket")`) the result is identical — the delay survives `qiskit_to_qasm3` and `transform_notation` intact and dies at the last hop.

### Expected behavior

A `delay` targeting a Braket device should raise (or at minimum emit a loud warning), the way we already do for barriers on the Cirq edge:

https://github.com/qBraid/qbraid/blob/a51d20ba180b6de066d93b10f5bd8c70e0e46f6a/qbraid/transpiler/conversions/qasm3/qasm3_to_cirq.py#L77-L82

Silently producing a semantically different circuit is the worst available outcome — worse than failing, because the user gets data back.

### Actual behavior

The delay is dropped with no diagnostic. The drop happens inside the Braket SDK: `Circuit.from_ir` dispatches via `@singledispatchmethod visit` and has **no** `visit.register` for `DelayInstruction`, so the node falls through to the generic visitor, gets walked, and contributes zero instructions.

We hand Braket valid QASM — neither pyqasm nor `transform_notation` strips the delay — so this is Braket SDK behavior we're silently inheriting at:

https://github.com/qBraid/qbraid/blob/a51d20ba180b6de066d93b10f5bd8c70e0e46f6a/qbraid/transpiler/conversions/qasm3/qasm3_to_braket.py#L80-L101

Worth noting this isn't a fixable gap on Braket's side: Braket has **no gate-level delay at all**. `braket.circuits.Circuit` has no `.delay()` method; the only delay is `PulseSequence.delay()`, at the pulse level. [Braket's supported-statements doc](https://docs.aws.amazon.com/braket/latest/developerguide/braket-openqasm-supported-features.html) lists `delay[...]` only inside `cal`/`defcal` blocks applied to *frames*, never to qubits. So raising is the correct behavior — there's nothing to lower to.

### Related: verbatim boxes are dropped on the same path

Found while reproducing the customer's "I tried verbatim and nothing changed":

```python
qasm_v = """OPENQASM 3.0;
bit[1] b;
#pragma braket verbatim
box{
rx(1.5707963267948966) $0;
delay[100000ns] $0;
rx(1.5707963267948966) $0;
}
b[0] = measure $0;
"""
cv = Circuit.from_ir(source=qasm_v)
[type(i.operator).__name__ for i in cv.instructions]   # ['Rx', 'Rx', 'Measure']
# no StartVerbatimBox / EndVerbatimBox compiler directives at all
```

The verbatim box is lost entirely, so a user asking for verbatim compilation via QASM silently doesn't get it. (It survives fine when built natively with `Circuit.add_verbatim_box()` — it's specifically the QASM import path.) Possibly worth splitting into its own issue.

### The inconsistency

Timing constructs currently behave three different ways through `qasm3_to_braket`:

| construct | outcome |
|---|---|
| `delay` | silently dropped |
| `barrier` | silently dropped |
| `box` | box dropped, **contents still applied** |
| `stretch` | raises `NameError` → `QasmError` |
| `duration` | raises `AttributeError` → `QasmError` |

`stretch` and `duration` raise only incidentally, via the catch-all `except Exception`. The two that users actually reach for are the two that fail silently.

For contrast, every other target handles delay correctly:

- **pyquil / Rigetti QCS** — translated to a real Quil `DELAY`: https://github.com/qBraid/qbraid/blob/a51d20ba180b6de066d93b10f5bd8c70e0e46f6a/qbraid/transpiler/conversions/openqasm3/openqasm3_to_pyquil.py#L210-L212
- **cirq** — raises `QasmError`
- **native qBraid `qasm3` spec** — preserved in the payload

### Suggested fix

Add a delay/timing check in `qasm3_to_braket` before handing off to `Circuit.from_ir`, raising `ProgramConversionError` with a message pointing at the two real options (Braket Pulse via `PulseSequence.delay`, or Rigetti direct via Quil-T `DELAY`). Mirroring the existing barrier-warning precedent on the Cirq edge would keep this consistent with how we already treat instructions we can't lower.

A `validate=` lambda on `ProgramSpec(Circuit)` at [`aws/provider.py#L181`](https://github.com/qBraid/qbraid/blob/a51d20ba180b6de066d93b10f5bd8c70e0e46f6a/qbraid/runtime/aws/provider.py#L181) is an alternative, but it catches the problem later and only for the AWS runtime path — the transpiler is where the information is actually lost.

Separately, there's no `supportedOperations` validation anywhere in the AWS runtime (`grep -rn "supportedOperations" qbraid/` returns nothing), which is the more general reason this class of bug can reach a QPU unchallenged.

### Customer workaround (documented here for reference)

- **Rigetti direct via qBraid** — Quil-T `DELAY 0 <t>`, with `device.set_options(transform=False)` to skip quilc (it rejects Quil-T) and `runtime_options={"prepend_default_calibrations": True}`. This is the path Rigetti's own [T1 tutorial](https://pyquil-docs.rigetti.com/en/stable/quilt_getting_started.html) uses.
- **Braket Pulse on Cepheus** — `Circuit().pulse_gate(q, PulseSequence().delay(frame, t))`, per the AWS [dynamical decoupling post](https://aws.amazon.com/blogs/quantum-computing/suppressing-errors-with-dynamical-decoupling-using-pulse-control-on-amazon-braket/). Frame names vary by device generation, so `device.frames` must be introspected.

rryoung98:
Fixes #1269

## Problem

`delay` instructions were **silently discarded** when converting to Braket — no exception, no warning, no log line. Circuits that depend on idle time (T1/T2 characterization, dynamical decoupling) ran as zero-delay circuits and returned plausible-looking but meaningless data.

This came in as a customer report: they were sweeping delay to measure T1 on Rigetti `Cepheus-1-108Q` via AWS and got a flat curve. The same code works on IBM. Their delays never reached the QPU — every point in the sweep was the same zero-delay circuit, and they burned QPU time before noticing.

Reproduced against `amazon-braket-sdk` 1.91.0:

```python
from braket.circuits import Circuit
c = Circuit.from_ir(source="""OPENQASM 3.0;
bit[1] b;
qubit[1] q;
h q[0];
delay[100000ns] q[0];
b[0] = measure q[0];
""")
len(c.instructions)   # 2 -- H and Measure. The delay is gone.
```

## Why raise rather than warn-and-drop

Braket has **no gate-level delay at all**. `braket.circuits.Circuit` has no `.delay()` method; the only delay is `PulseSequence.delay()`, at the pulse level. [Braket's supported-statements doc](https://docs.aws.amazon.com/braket/latest/developerguide/braket-openqasm-supported-features.html) lists `delay[...]` only inside `cal`/`defcal` blocks applied to *frames*, never to qubits. So there is nothing to lower a delay to.

We hand Braket valid QASM — neither pyqasm nor `transform_notation` strips the delay. The drop happens inside `Circuit.from_ir`, which dispatches via `@singledispatchmethod visit` and has no `visit.register` for `DelayInstruction`; the node falls through to the generic visitor and contributes zero instructions.

This differs from the `remove_barriers()` precedent on the Cirq edge, which warns and drops. Dropping a barrier loses a compiler hint; dropping a delay returns a circuit that runs and measures the wrong thing. Failing is the better outcome.

## Change

- `contains_delay(qasm3)` in `qasm3_to_braket.py`, using an `openqasm3` visitor so nested delays (e.g. inside a `box`) are caught too.
- `qasm3_to_braket` raises `QasmError` up front, naming the two ways to actually get idle time on hardware: Braket Pulse (`PulseSequence.delay`), or Rigetti direct via Quil-T `DELAY`.
- Corrected the docstring's `Raises:` — it claimed `ProgramConversionError`, but the function raises `QasmError`.

The check only parses when the reserved word `delay` appears in the source, so the common path costs one substring scan rather than a full parse.

The customer's exact scenario now fails loudly instead of silently:

```
ProgramConversionError: Failed to convert 'qiskit' to 'braket' ...
QasmError: Delay instructions are not supported by Amazon Braket, which has no
gate-level delay: braket.circuits.Circuit cannot represent one, and Braket
discards delays during conversion without raising. Express idle time at the
pulse level with braket.pulse.PulseSequence.delay, or run the program on
Rigetti directly, where Quil-T's DELAY is supported.
```

## Tests

Added to `tests/transpiler/braket/test_braket_qasm3.py`:

- delay raises across units (`ns`/`dt`/`us`) and for a whole-register target
- delay nested in a verbatim box raises (verbatim is not a workaround — the box is discarded on this path too)
- the error names both alternatives
- programs without a delay convert exactly as before
- the keyword prefilter is a fast path only: an identifier like `delay_reg` does not trip the check
- malformed input defers to the downstream parsers, which report it in context

## Notes for the reviewer

- **Not addressed here:** `barrier` and `box` are also silently dropped on this path (`box` drops the box but still applies its contents), and the verbatim box is lost through `from_ir` even though it survives when built natively via `Circuit.add_verbatim_box()`. These are noted in #1269 and may deserve their own issue — this PR is scoped to the reported delay bug.
- **Two pre-existing test failures** show up locally with newer pinned deps and are unrelated to this change (both fail identically on clean `main`): `test_qasm3_to_braket_error_includes_detail` (braket 1.124 handles `c3x` differently than 1.91) and `test_braket_coverage[qiskit-1.0]`.
- For contrast, every other target already handles delay correctly: pyquil/Rigetti translates it to a real Quil `DELAY`, cirq raises, and the native qBraid `qasm3` spec preserves it in the payload.


ryanhill1:
@rryoung98 I pushed a small refactor (72c0abf) to the delay detection — behavior and error message are unchanged, all 17 tests in the file still pass. Wanted to explain the reasoning:

**What changed:** delay detection now runs on the AST that `pyqasm.loads()` already parses (`module.original_program`) instead of a separate `openqasm3.parse()` up front.

**Why:** `qasm3_to_braket` already calls `pyqasm.loads(qasm)`, and `Circuit.from_ir` parses again — so `contains_delay`'s own `openqasm3.parse` was a third parse of the same source. `openqasm3.parse` is expensive (I measured ~15 ms on a ~200-line program, ~150 ms at ~2000 lines), so this avoids it entirely.

The key detail that makes it work for the verbatim-box case: `pyqasm.loads()` **succeeds** even on a `#pragma braket verbatim box{...}` — it's `unroll()`/`validate()` that raise `ValidationError`, not `loads()`. And `original_program` is populated at load time regardless. So I split `loads` out of the `unroll` try block, run the visitor on `original_program` before unrolling, and it still catches a delay nested in a verbatim box (that test now guards this behavior). Confirmed empirically across basic / whole-register / verbatim-box / `delay_reg`-identifier / commented-out-delay cases.

Also considered a regex bridge but ruled it out: it's actually *slower* than the current substring-guarded path on the common no-delay case (it must scan the whole string vs. a fast `in` check) and it false-positives on a commented-out `// delay[...]`.

Two small follow-ups I reworked along the way:
- The two `contains_delay(str)` unit tests became behavioral tests on `qasm3_to_braket` (identifier `delay_reg` converts to `Circuit().x(0)`; malformed input raises a non-delay `QasmError`).
- Longer term this detection belongs in `pyqasm` as a `has_delays()` method (alongside `has_barriers`/`has_measurements`) — with the caveat that those iterate only top-level statements, so a `has_delays` there should recurse to match the box behavior. Happy to open that issue/PR.

PR #1284, opened 2026-07-22
fix: support object arguments in cached methods

### LINKED ISSUE #1266: cached_method raises TypeError for non-JSON-serializable arguments (e.g. get_devices(aws_session=...))
## Description

`cached_method`'s key generation (`qbraid._caching._generate_cache_key`) serializes the decorated method's arguments with `json.dumps`, so calling a cached method with any non-JSON-serializable argument raises:

```
TypeError: Object of type AwsSession is not JSON serializable
```

This is reachable from public API today, e.g.:

```python
from braket.aws import AwsSession

provider = BraketProvider()
provider.get_devices(aws_session=AwsSession())  # TypeError from json.dumps
```

`BraketProvider.get_devices` documents an `aws_session` parameter, but passing one crashes in cache-key generation before the method body runs. Any other object argument (device instances, `datetime`, custom filters) hits the same wall.

## Scope / history

Pre-existing — this is **not** a regression from #1244 (reproduced identically on `qbraid==0.12.0`, where `_generate_cache_key` already used `json.dumps` before delegating to the inner `lru_cache`). #1244 fixed the sibling limitation (unhashable-but-serializable args like `list`/`dict`); this issue covers the remaining case of serializable-by-neither args.

Workarounds until fixed: `DISABLE_CACHE=1`, or the `cache_disabled(instance)` context manager (works correctly as of #1244).

## Proposal

Make key generation total by falling back to `repr` for objects `json` can't encode:

```python
key_str = json.dumps(key_data, sort_keys=True, default=repr)
```

Trade-off to weigh in review: `default=repr` means two objects with the same repr share a cache entry, and objects whose repr embeds a memory address (`<... object at 0x104f3d550>`) get per-object entries — both acceptable for a 120s-TTL cache, but worth a comment and a test. An alternative is to fall back to `id(obj)` for unencodable objects, mirroring the `hash(instance)` → `id(instance)` fallback used for instance identity.

Sanjays2402:
Closes #1266

## Summary of changes

Fall back to `repr` when cache-key arguments are not JSON serializable, allowing documented object arguments to reach cached methods. Adds a regression test and changelog entry.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved caching behavior for calls with arguments that cannot be JSON-serialized.
  * Such calls now execute successfully without being stored in the cache, avoiding potentially conflicting cache keys.
  * Added handling for circular and excessively nested arguments.

* **Documentation**
  * Updated the unreleased changelog to document the new cache-bypass behavior.


qbraid/_caching.py
```diff
@@ -60,18 +60,19 @@ def _generate_cache_key(instance: Any, func_name: str, args: tuple, kwargs: dict
         "args": args,
         "kwargs": kwargs,
     }
-    key_str = json.dumps(key_data, sort_keys=True)
+    key_str = json.dumps(key_data, sort_keys=True, default=repr)
```
ryanhill1: `default=repr` trades the reported crash for silent wrong results. Two failure modes, both verified locally:

1. **Freed-address reuse -> stale hits.** Objects without a custom repr key as `<AwsSession object at 0x...>`. Once the object is gc'd, CPython readily reuses the address (observed on the first attempt: `x = object(); r1 = repr(x); del x; y = object(); repr(y) == r1` -> `True`). So `get_devices(aws_session=s1)` caches, `s1` is collected, a differently-configured `s2` lands on the same address, and within the 120s TTL the method silently returns `s1`'s results for `s2` -- the same bug class #1244 fixed for instance identity, reintroduced for arguments.
2. **Truncated-repr collisions.** Two different 10,000-element numpy arrays (one element changed) have byte-identical truncated reprs -> identical keys -> wrong cached result.

Also, key generation still isn't total: `default=` only applies to values, so dicts with tuple keys (`TypeError: keys must be str...`) or mixed-type keys under `sort_keys` still crash before the method body.

To be fair, this PR faithfully implements what #1266 proposed -- that's on the issue, not on you: the issue (mine) modeled address-bearing reprs as merely producing per-object entries, missing the reuse-after-gc mode, and its "make key generation total" claim doesn't hold for the dict-key cases.

**Requested change:** bypass the cache instead of guessing a key -- a cache miss is always correct, a wrong hit never is:

```python
try:
    key = _generate_cache_key(self, func.__name__, args, kwargs)  # strict json.dumps
except TypeError:
    return func(self, *args, **kwargs)  # unkeyable args -> call through, uncached
```

This fixes the reported crash, absorbs the dict-key crashes for free, and can never serve a wrong result; the only cost is no caching for exotic arguments. The regression test then inverts (`call_count == 2`, asserting the bypass), plus keep one test that JSON-serializable args still cache.

ryanhill1 (changes requested):

qbraid/_caching.py
Sanjays2402: You're right on both counts, and the address-reuse one is the killer — I'd been thinking of `id`/`hash` as a safe identity component, but neither survives gc, so my follow-up commit had the same flaw wearing a different hat. Pushed f6c1e7a implementing the bypass.

`_generate_cache_key` is now a strict `json.dumps` with no `default=`, and the wrapper does:

```python
try:
    key = _generate_cache_key(self, func.__name__, args, kwargs)
except TypeError:
    return func(self, *args, **kwargs)
```

That also covers the dict-key cases you mentioned, since `default=` only ever applied to values.

Tests inverted as you described: `test_cached_method_bypasses_cache_for_non_json_serializable_args` asserts `call_count == 2` after two identical object calls plus `currsize == 0`, and includes a `{(1, 2): "a"}` argument to cover the dict-key path. `test_generate_cache_key_rejects_unserializable_args` asserts the key generator itself raises for both shapes. The JSON-serializable caching tests (including the unhashable `list` arg one from #1244) are unchanged and still pass. `pytest tests/top_level/test_caching.py` is 11 passed; reverting `_caching.py` fails exactly the two new tests.

Changelog entry updated to describe the bypass rather than the repr fallback.

qbraid/_caching.py
```diff
@@ -104,7 +111,13 @@ def wrapper(self, *args, **kwargs):
                 or maxsize == 0
             ):
                 return func(self, *args, **kwargs)
-            key = _generate_cache_key(self, func.__name__, args, kwargs)
+            try:
+                key = _generate_cache_key(self, func.__name__, args, kwargs)
+            except TypeError:
```
ryanhill1: `TypeError` misses the two other ways `json.dumps` fails, so these still crash before the method body:

```
circular ref (a = []; a.append(a)):  ValueError: Circular reference detected
deeply nested list:                  RecursionError: maximum recursion depth exceeded
```

Exotic, but the docstring now claims anything JSON can't encode bypasses the cache — worth making that true:

```suggestion
            except (TypeError, ValueError, RecursionError):
```

ryanhill1 (commented):
Bypass is implemented as requested and both failure modes are gone — verified locally: the freed-address and truncated-numpy-repr cases now recompute instead of serving a stale hit, the tuple-key and mixed-key dicts no longer crash, and the `list` arg from #1244 still caches. 11 passed.

Two minor things and this is good to go: the exception widening inline, and a rebase — `CHANGELOG.md` is the only conflict against `main`.

qbraid/_caching.py
Sanjays2402: Taken, in 51dd213 — the guard is now `except (TypeError, ValueError, RecursionError)`, and both the `_generate_cache_key` and `cached_method` docstrings name the three failure modes instead of implying only `TypeError`.

Extended both tests with your two cases: `test_generate_cache_key_rejects_unserializable_args` asserts the raise for a self-referential list and for one nested `sys.getrecursionlimit() + 100` deep, and `test_cached_method_bypasses_cache_for_non_json_serializable_args` now drives both through the decorator, asserting `call_count` still increments and `currsize == 0`. Reverting just the tuple back to `TypeError` fails that second test with the raw `ValueError: Circular reference detected` from `json/encoder.py`, so it is covering the branch. `tests/top_level/test_caching.py` is 11 passed.

Also merged main to clear the CHANGELOG conflict.

ryanhill1 (approved):

PR #1291, opened 2026-07-23
Clarify sparse pyQuil unitary errors

### LINKED ISSUE #1287: PyQuilProgram.unitary() fails with an opaque error on sparse qubit indices
## Summary

`PyQuilProgram._unitary()` fails with a confusing, low-level error for any pyQuil program whose qubit indices are sparse (e.g. a hardware-mapped program addressing `$4` and `$13`). The workaround exists and works, but it is undiscoverable from the error.

## Reproduction

```python
from qbraid.programs import load_program
from pyquil import Program
from pyquil.gates import X, CZ

p = load_program(Program(X(0), CZ(0, 2)))
print(p.qubits, p.num_qubits)   # [0, 2] 2
p.unitary()                     # ValueError: Permutation SWAP index not valid
```

Via `circuits_allclose`, the same cause surfaces as a different error:

```python
circuits_allclose(Program(X(4), CZ(4, 13)), Program(X(4), CZ(4, 13)), strict_gphase=True)
# IndexError: index 0 is out of bounds for axis 0 with size 0
```

## Cause

`qbraid/programs/gate_model/pyquil.py:107`:

```python
return program_unitary(program_copy, n_qubits=self.num_qubits)
```

`num_qubits` is the *count* of distinct qubits (`len(self.qubits)`), but `program_unitary` needs an `n_qubits` that bounds the largest qubit *index*. For contiguous programs these coincide, so the bug is invisible; for sparse ones `n_qubits` is too small and pyQuil fails somewhere inside its permutation logic.

## Workaround (works today)

`circuits_allclose(..., index_contig=True)` calls `remove_idle_qubits()` on both programs first (`qbraid/interface/circuit_equality.py:132-134`), which compacts indices and makes the comparison succeed. `PyQuilProgram.remove_idle_qubits()` is already implemented and correct:

```python
p = load_program(Program(X(0), CZ(0, 2)))
p.remove_idle_qubits()
p.unitary().shape   # (4, 4)
```

So nothing is fundamentally broken. The problem is purely that the failure mode is opaque: neither error message mentions qubit indexing or points at `index_contig`.

## Suggested fix

Either would do; the second is the smaller change.

1. Have `PyQuilProgram._unitary()` size the operator by `max(self.qubits) + 1` rather than `len(self.qubits)`, so sparse programs produce a correct (if larger) unitary with no caller change. Note this makes the operator exponential in the largest index, so a program touching `$108` is not representable in memory. That argues for (2).
2. Raise a clear error from `_unitary()` when `max(self.qubits) >= self.num_qubits`, naming the sparse indices and pointing at `index_contig=True`.

## Why this is worth fixing

Physical-qubit programs are now first-class on the OpenQASM 3 to pyQuil path (#1286), and they are exactly the programs that are sparse: a circuit hand-mapped to a Rigetti lattice will routinely address qubits like `$13`, `$27`, `$38`. Anyone reaching for `circuits_allclose` to check such a conversion hits this with no indication of what went wrong or that a supported option already solves it.

One caveat worth recording for whoever picks this up: `index_contig=True` renumbers both operands, so it verifies gate structure but is *blind to absolute qubit indices*. It returns `True` when comparing `X 4; CZ 4 13` against `X 0; CZ 0 1`. It is therefore the right tool for semantic equivalence but the wrong one for asserting that a hardware mapping was preserved. Tests that care about placement need to compare emitted Quil instead (see `test_transpile_physical_qubits_to_pyquil` in #1286).

Sanjays2402:
Closes #1287

## Summary of changes

Sparse pyQuil programs now fail before pyQuil's low-level permutation logic and explain both supported qubit-compaction options. The regression test fails with `Permutation SWAP index not valid` before the fix and passes with the actionable error; all 8 focused pyQuil program tests and Ruff checks pass.

### DIFF
--- a/qbraid/programs/gate_model/pyquil.py
+++ b/qbraid/programs/gate_model/pyquil.py
@@ -104,6 +104,13 @@ class PyQuilProgram(GateModelProgram):
     def _unitary(self) -> np.ndarray:
         """Return the unitary of a pyQuil program."""
         program_copy = self.remove_measurements(self.program)
+        qubits = sorted(program_copy.get_qubits())
+        if qubits and qubits != list(range(len(qubits))):
+            raise ValueError(
+                "PyQuilProgram.unitary() requires contiguous qubit indices starting at 0; "
+                f"got {qubits}. Call remove_idle_qubits() first or use "
+                "circuits_allclose(..., index_contig=True)."
+            )
         return program_unitary(program_copy, n_qubits=self.num_qubits)
 
     def serialize(self) -> Program:
--- a/qbraid/programs/gate_model/pyquil.py
+++ b/qbraid/programs/gate_model/pyquil.py
@@ -105,7 +105,7 @@ class PyQuilProgram(GateModelProgram):
         """Return the unitary of a pyQuil program."""
         program_copy = self.remove_measurements(self.program)
         qubits = sorted(program_copy.get_qubits())
-        if qubits and qubits != list(range(len(qubits))):
+        if qubits and max(qubits) >= self.num_qubits:
             raise ValueError(
                 "PyQuilProgram.unitary() requires contiguous qubit indices starting at 0; "
                 f"got {qubits}. Call remove_idle_qubits() first or use "
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Writing an entry:
 ### Removed
 
 ### Fixed
+- Sparse-indexed pyQuil programs now raise an actionable unitary error that points to the supported qubit-compaction options instead of leaking a low-level permutation failure.
 - Fixed `openqasm3_to_pyquil` raising `AttributeError` for programs addressing physical qubits (`x $0;`) rather than a declared register. Physical qubit indices now pass through verbatim and are not identity-padded, since such a program is already mapped to specific hardware ([#1286](https://github.com/qBraid/qBraid/pull/1286))
 - Fixed OpenQuantum `prepare_job` / `get_preparation_result` / `create_job` failing without a clear message when the API rejected the request. Responses are now parsed for the error `type`, mapping `TERMS_OF_USE_REQUIRED` to a `QbraidRuntimeError` pointing at https://www.openquantum.com, and other failures to a message including HTTP status, error type, and API detail ([#1282](https://github.com/qBraid/qBraid/pull/1282))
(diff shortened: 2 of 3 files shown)


qbraid/programs/gate_model/pyquil.py
```diff
@@ -104,6 +104,13 @@ def remove_measurements(original_program):
     def _unitary(self) -> np.ndarray:
         """Return the unitary of a pyQuil program."""
         program_copy = self.remove_measurements(self.program)
+        qubits = sorted(program_copy.get_qubits())
+        if qubits and qubits != list(range(len(qubits))):
```
ryanhill1: This checks the *gate* qubits for contiguity, but `program_unitary` only needs the largest gate index bounded by the `n_qubits` you pass — and that's `self.num_qubits`, which counts measurement qubits too. The two diverge whenever a program measures a qubit it never gates:

```python
prog = Program(Declare("ro", "BIT", 3), H(0), CZ(0, 2))
for q in range(3):
    prog += MEASURE(q, ("ro", q))
load_program(prog).unitary()
```

`num_qubits` is 3 here, so the call is correctly sized and returns `(8, 8)` on `main`. But `remove_measurements` strips the MEASUREs first, leaving gate qubits `[0, 2]`, so this now raises — and it's a dead end: `get_qubits()` is already `[0, 1, 2]`, so `remove_idle_qubits()` and `index_contig=True` are both no-ops that re-raise the identical error.

The condition from #1287 fixes it and keeps your new test green as-is:

```suggestion
        if qubits and max(qubits) >= self.num_qubits:
```

Worth adding a regression test for the measure-the-whole-register shape too, since that's the case the current guard gets wrong.

ryanhill1 (changes requested):
Verified both symptoms from #1287 are resolved — the sparse `unitary()` call and the `circuits_allclose` path now give the actionable error, and both suggested remedies work. One issue with the guard condition inline; also needs a rebase, `main` has moved and the CHANGELOG conflicts.

qbraid/programs/gate_model/pyquil.py
Sanjays2402: You are right, and your example does raise on the branch as it stood. Switched to the #1287 condition:

    if qubits and max(qubits) >= self.num_qubits:

and added `test_unitary_allows_measured_only_qubits` covering the measure-the-whole-register shape (`Declare ro BIT 3`, `H 0`, `CZ 0 2`, MEASURE 0-2) asserting an (8, 8) unitary. With the old `qubits != list(range(len(qubits)))` condition restored that test fails, and it passes with the fix. The sparse test is unchanged and still green. 9 passed in tests/programs/test_program_circuits_pyquil.py. Pushed as e0e706b.

ryanhill1 (approved):

PR #1297, opened 2026-07-28
fix(runtime): make the Quantinuum provider production-ready

rryoung98:
## Summary

Production-readiness pass on the Quantinuum provider, covering the transport layer and the job class. Started as the compile-wait bound; review showed that bound was necessary but nowhere near sufficient, and probing the job path turned up three further defects that would have shipped.

## Why the compile wait was unbounded at all

`qnx.jobs.wait_for` has not always been unbounded. On our declared floor, qnexus 0.39.0, the signature is `timeout: float | None = 900.0`. By 0.43 that default had become `None`. Upstream silently dropped a 900s bound, which is why this surfaced now. The fix works on 0.39.0 too, so no floor break.

## Transport (`_transport.py`, new)

**A per-call timeout does not bound the thread.** qnexus builds its process-wide client with `httpx.Client(..., timeout=None)`, and `submit` makes six HTTP calls. Bounding only `wait_for` leaves the other five able to pin the calling thread on a hung socket — the exact leak the original summary claimed to fix. A retry helper cannot cover it either: a hang raises nothing. `ensure_bounded_client()` sets a per-request timeout on the shared client (`QUANTINUUM_NEXUS_HTTP_TIMEOUT`, default 60s) and is applied at every qnexus entry point in both the device and job classes. It is a per-request bound, not an overall deadline, so it does not cap job duration: `wait_for` waits over a websocket and falls back to short polls, both well inside it.

**The retry missed the likeliest failure.** qnexus checks status codes by hand and raises its own exception types, so a NEXUS 502/503 never arrives as an `httpx.TransportError`. Retries now cover `ResourceCreateFailed` / `ResourceFetchFailed` gated on gateway-class codes, leaving 4xx to surface immediately. Backoff is halved and jittered — sleeping holds the thread, and the failure being guarded against hits every concurrent submit at once, so a long lockstep sleep adds pool pressure in exactly the scenario it is meant to survive.

Also: `CompilationResultRef.get_output` is lazy and issues its own request, so the whole result-fetch stage moved inside the retry; `wait_for`'s `qnx_exc.JobError` (compile errored or cancelled) is wrapped rather than escaping bare; `retry_transient(attempts=0)` raised a bare `AssertionError` without calling the function and now rejects the argument; and `COMPILE_TIMEOUT` / `HTTP_TIMEOUT` / `OPT_LEVEL` are validated, so `COMPILE_TIMEOUT=0` no longer makes every submit report "did not complete within 0 seconds" — which reads like a NEXUS outage rather than a config error.

## Job correctness

**`status()` could never observe a job finishing.** It read `ref.last_status`, a plain pydantic field fixed when the reference is built, and `_get_ref()` returns the cached reference. So `job = device.run(circuit)` followed by `job.result()` polled the submission-time status until the caller timed out. `qbraid-runtime-api` happened to dodge this by constructing a fresh `QuantinuumJob` per poll; the direct-SDK path did not. Status now comes from `qnx.jobs.status(ref)`, which also carries the timestamps and error detail, so `execution_time_s()` and the failure message stop reading the snapshot too.

**The status map covered 6 of `JobStatusEnum`'s 10 members** and included an `INITIALIZING` key NEXUS never emits. The costly omissions were `TERMINATED` and `DEPLETED` (credits exhausted): both stop the job, both fell through to `UNKNOWN`, and `UNKNOWN` is not terminal — so a stopped job was polled forever and, downstream, its escrow never settled. `SUBMITTED` (the first status most jobs report), `CANCELLING` and `RETRYING` were unmapped as well.

**`simulator="E" in device_id.upper()` matches `Helios-1`,** so Quantinuum's QPU would be profiled as a simulator. Quantinuum derives non-hardware targets by suffix (`E`, `LE`, `SC`), so the check now anchors at the end of the name.

## Tests

The tests that should have caught the two job defects instead certified them. `test_status_map_covers_expected_states` asserted the map equalled a hand-written copy of itself, staying green no matter which states were missing; every status test built a `SimpleNamespace(last_status=...)`, which is the stale field the code was wrongly reading; and the counts fixtures were palindromic (`{"00", "11"}`), so a bit-order flip was undetectable.

Now: the map is checked against `JobStatusEnum` itself; the status tests drive a real `qnx.jobs.status`, one walking a ref that stays `SUBMITTED` while NEXUS progresses to `COMPLETED`; result fixtures use asymmetric bitstrings; and the compile-timeout mapping is pinned by driving the real `wait_for` with a strategy that never returns, rather than fabricating the exception and asserting the mapping of the fabrication.

81 tests, pylint 10.00/10. `CHANGELOG.md` updated under `[Unreleased]`.

## Not in this PR

The `qnx.projects.get_or_create` TOCTOU (two concurrent first-time submits can create duplicate projects) is real but sits in qnexus's API shape; worth its own change.


rryoung98:
Added a second hardening fix from the same stress campaign: transient NEXUS connection errors (`Server disconnected`, `Connection reset by peer` — ~12% of submits under 40-job concurrent load) are now retried with backoff in the pre-execute stages only. `start_execute_job` is deliberately not retried to avoid double-submission.

qbraid/runtime/quantinuum/device.py
```diff
@@ -40,6 +42,33 @@ class QuantinuumDeviceError(QbraidRuntimeError):
...
+    read-only fetches). Never wrap ``start_execute_job`` in this: a disconnect
+    after the server accepted the request would double-submit the job.
+    """
+    # pylint: disable-next=import-outside-toplevel
+    import httpx
+
+    for attempt in range(attempts):
+        try:
+            return fn()
+        except (httpx.TransportError, ConnectionError) as err:
```
ryanhill1: **Blocker: a NEXUS 5xx is never retried — the most likely transient failure slips through.**

`httpx.HTTPStatusError` isn't a `TransportError` subclass, and qnexus never raises it anyway: it checks status codes by hand and raises its own types. From `qnx.circuits.upload`:

```python
res = get_nexus_client().post("/api/circuits/v1beta2", json=req_dict)
if res.status_code != 201:
    raise qnx_exc.ResourceCreateFailed(message=res.text, status_code=res.status_code)
```

So this `except` only ever fires on connection drops. I measured it directly against the branch:

```
NEXUS 5xx (ResourceCreateFailed)          -> attempts = 1  (retried? False)
dropped connection (RemoteProtocolError)  -> attempts = 3  (retried? True)
```

For a cloud API, 502/503 is the more common blip. Suggest adding the qnexus types and gating on the status code, e.g.:

```python
import qnexus.exceptions as qnx_exc

RETRYABLE_STATUS = {502, 503, 504}
...
except (httpx.TransportError, ConnectionError) as err:
    ...
except qnx_exc.ResourceCreateFailed as err:
    if err.status_code not in RETRYABLE_STATUS:
        raise
    ...
```

Worth checking which `qnx_exc` types the other wrapped calls (`projects.get_or_create`, `jobs.results`) raise — they follow the same hand-rolled pattern.

qbraid/runtime/quantinuum/device.py
```diff
@@ -40,6 +42,33 @@ class QuantinuumDeviceError(QbraidRuntimeError):
...
+            if attempt == attempts - 1:
+                raise
+            delay = base_delay * (2**attempt)
+            logger.warning("Transient NEXUS connection error (%s); retrying in %.1fs", err, delay)
+            time.sleep(delay)
```
ryanhill1: This works against the PR's own motivation. The summary describes 40 concurrent submits exhausting a shared executor; `time.sleep` holds the thread for 1s then 2s per wrapped call. During a NEXUS blip that hits all 40, every thread now sleeps rather than failing fast, so pool pressure goes **up** in exactly the scenario this PR is fixing.

Combined with the 900s compile bound, a single submit can hold a thread for ~15 minutes. Is the executor in `qbraid-runtime-api` sized for that? If it's the same pool the status polls use, a 900s default may just relocate the starvation rather than remove it — a shorter default plus a client-level timeout might serve better.

qbraid/runtime/quantinuum/device.py
```diff
@@ -40,6 +42,33 @@ class QuantinuumDeviceError(QbraidRuntimeError):
...
+    raise AssertionError("unreachable")  # pragma: no cover
```
ryanhill1: Minor: this is reachable, not unreachable — `_retry_transient(fn, attempts=0)` skips the loop entirely and raises `AssertionError` without ever calling `fn`. No current caller passes 0, but the `# pragma: no cover` will hide it if one ever does.

Either guard the input (`if attempts < 1: raise ValueError(...)`) or restructure so the final attempt is outside the loop.

qbraid/runtime/quantinuum/device.py
```diff
@@ -125,30 +158,54 @@ def submit(  # pylint: disable=arguments-differ
...
+                name=unique("compile"),
+                optimisation_level=resolved_opt_level,
+                backend_config=backend_config,
+                project=project,
+            )
+        )
+        # NOTE: blocking wait during dispatch; compilation time depends on queue and program
+        # size. The wait is bounded: an unbounded wait_for leaks the calling thread forever
+        # if the NEXUS compile job hangs, which starves thread pools in server deployments.
+        compile_timeout = float(
```
ryanhill1: Per the [Fail Loudly on Missing Data](https://github.com/qBraid/qBraid/blob/main/CONTRIBUTING.md#code-style) guidance, this needs validating — an operator typo silently breaks every submission. Probed on the branch:

```
'abc' -> ValueError: could not convert string to float: 'abc'   (doesn't name the variable)
'15m' -> ValueError: could not convert string to float: '15m'
'0'   -> 0.0    accepted; asyncio.wait_for expires immediately
'-5'  -> -5.0   accepted; same
```

`QUANTINUUM_NEXUS_COMPILE_TIMEOUT=0` makes every submit fail with "did not complete within 0 seconds", which reads like a NEXUS outage rather than a config error. Suggest:

```python
raw = os.getenv("QUANTINUUM_NEXUS_COMPILE_TIMEOUT")
if raw is None:
    compile_timeout = float(DEFAULT_COMPILE_TIMEOUT_SECONDS)
else:
    try:
        compile_timeout = float(raw)
    except ValueError as err:
        raise QuantinuumDeviceError(
            f"QUANTINUUM_NEXUS_COMPILE_TIMEOUT must be a number of seconds, got {raw!r}."
        ) from err
    if compile_timeout <= 0:
        raise QuantinuumDeviceError(
            f"QUANTINUUM_NEXUS_COMPILE_TIMEOUT must be positive, got {compile_timeout}."
        )
```

qbraid/runtime/quantinuum/device.py
```diff
@@ -125,30 +158,54 @@ def submit(  # pylint: disable=arguments-differ
...
+            os.getenv("QUANTINUUM_NEXUS_COMPILE_TIMEOUT", str(DEFAULT_COMPILE_TIMEOUT_SECONDS))
         )
-        # NOTE: blocking wait during dispatch; compilation time depends on queue and program size.
         logger.info("Waiting for Quantinuum compilation job %s to complete...", compile_job.id)
-        qnx.jobs.wait_for(compile_job)
-        compiled_refs = [item.get_output() for item in qnx.jobs.results(compile_job)]
+        try:
+            qnx.jobs.wait_for(compile_job, timeout=compile_timeout)
```
ryanhill1, line 192: **Blocker: this bounds the poll loop, not the thread.**

qnexus builds its shared client with no timeout:

```python
_nexus_client = httpx.Client(base_url=CONFIG.url, auth=_auth_handler, timeout=None, ...)
```

`submit()` makes six qnexus HTTP calls and this bounds one of them. `projects.get_or_create` (L161), `circuits.upload` (L172), `start_compile_job` (L180), `jobs.results` (L204) and `start_execute_job` (L209) can each still block the calling thread forever on a hung socket — the precise leak the summary describes. `_retry_transient` can't cover it either: a hang raises nothing.

A client-level timeout would fix all six at once. Fine as a follow-up, but the PR description shouldn't claim the thread leak is resolved until it lands.

Separately: `wait_for` also raises `qnx_exc.JobError` when the compile job errors/is cancelled, which still escapes `submit()` unwrapped. Pre-existing, but it's the sibling of the case being fixed here — worth folding into the same `try` while you're in it.

tests/runtime/quantinuum/test_quantinuum_runtime.py
```diff
@@ -316,6 +320,121 @@ def test_submit_accepts_project_and_opt_level_kwargs(
...
+        # pylint: disable-next=import-outside-toplevel
+        import asyncio
+
+        # pylint: disable-next=import-outside-toplevel
+        from pytket import Circuit
+
+        mock_get_or_create.return_value = MagicMock(name="project")
+        mock_compile.return_value = MagicMock(id="compile-job-id")
+        mock_upload.side_effect = [MagicMock(name="circuit-ref")]
+        mock_wait.side_effect = asyncio.TimeoutError()
```
ryanhill1, line 438: This is a self-reinforcing test in the sense the new [Testing Philosophy](https://github.com/qBraid/qBraid/blob/main/CONTRIBUTING.md#testing-philosophy) section warns about: it fabricates the exception *and* asserts the mapping of that same fabricated exception, so it passes regardless of what qnexus actually raises. If upstream switched to a custom timeout type, the mapping would silently break and this stays green.

I checked independently and the assumption holds today — `wait_for` wraps the strategy in `asyncio.wait_for`, so a real expiry raises builtin `TimeoutError`, and `asyncio.TimeoutError is TimeoutError` on 3.11+. But nothing in the suite pins that.

Cheap to make real — drive the actual `wait_for` with a hanging strategy:

```python
class _HangingStrategy:
    async def get_status(self, job):
        await asyncio.sleep(60)

def test_qnexus_timeout_is_asyncio_timeout_error():
    """Pin the exception type QuantinuumDevice.submit maps to QuantinuumDeviceError."""
    with pytest.raises(asyncio.TimeoutError):
        qnx.jobs.wait_for(SimpleNamespace(id=uuid.uuid4()), timeout=0.2,
                          strategy=_HangingStrategy())
```

Runs in ~0.2s, no network, and fails loudly if qnexus changes the contract.

tests/runtime/quantinuum/test_quantinuum_runtime.py
```diff
@@ -316,6 +320,121 @@ def test_submit_accepts_project_and_opt_level_kwargs(
...
+        from pytket import Circuit
+
+        mock_get_or_create.return_value = MagicMock(name="project")
+        mock_compile.return_value = MagicMock(id="compile-job-id")
+        compiled_item = MagicMock()
+        compiled_item.get_output.return_value = MagicMock(name="compiled-ref")
+        mock_results.return_value = [compiled_item]
+        # First upload attempt drops the connection; the retry succeeds.
+        mock_upload.side_effect = [
+            httpx.RemoteProtocolError("Server disconnected without sending a response."),
```
ryanhill1, line 480: Same fabrication concern as the timeout test, though this one is lower risk: `qnx.circuits.upload` calls `get_nexus_client().post(...)` with no try/except, so raw httpx errors genuinely do escape and `RemoteProtocolError` is realistic.

The gap is what *isn't* covered — there's no test for a NEXUS 5xx, which is the case that silently isn't retried (see the `_retry_transient` comment). A test asserting the intended behaviour for `qnx_exc.ResourceCreateFailed(status_code=503)` would have caught it.

ryanhill1 (changes requested):
Directionally right and worth landing, but as it stands I don't think this stops the incident that motivated it. Two blockers below (the unbounded client, and the retry missing 5xx), plus a changelog entry that CONTRIBUTING now requires.

Reviewed against qnexus 0.43.1 and 0.45.0, plus the declared `>=0.39.0` floor. 48 tests pass, pylint 10.00/10.

### The premise checks out — and there's a detail worth adding

`qnx.jobs.wait_for` really is unbounded in current qnexus, but **it wasn't always**. In 0.39.0 (our declared floor) the signature is `timeout: float | None = 900.0`; by 0.43+ that default became `None`. Upstream silently dropped a 900s bound, which is almost certainly why this surfaced now. Worth putting in the PR body — it strengthens the case considerably. The fix works on 0.39.0 too, so no floor break.

I also verified the exception mapping against the real library rather than the mock: a genuine `wait_for` expiry raises builtin `TimeoutError`, and `asyncio.TimeoutError is TimeoutError` on 3.11+, so `except asyncio.TimeoutError` is correct. On 3.10 they're the same class from `asyncio.exceptions`, so it holds across our supported range.

### Blocker 1 — this doesn't actually bound the thread

qnexus constructs its shared client with no timeout at all:

```python
_nexus_client = httpx.Client(base_url=CONFIG.url, auth=_auth_handler, timeout=None, ...)
```

`submit()` makes six qnexus HTTP calls. This PR bounds one — the `wait_for` polling loop. `projects.get_or_create`, `circuits.upload`, `start_compile_job`, `jobs.results` and `start_execute_job` can each still block the calling thread forever on a hung socket, which is the exact failure described in the summary. `_retry_transient` doesn't help: a hang raises nothing, so there is nothing to retry.

The durable fix is a client-level timeout rather than a per-call one. Happy to open that as a follow-up if you'd rather keep this PR scoped — but then the description shouldn't claim the thread leak is resolved.

### Blocker 2 — the retry misses the most likely transient failure

Detail inline on `_retry_transient`. Short version: qnexus checks status codes by hand and raises its own exception types, so a NEXUS 502/503 is never retried. Measured:

```
NEXUS 5xx (ResourceCreateFailed)          -> attempts = 1  (retried? False)
dropped connection (RemoteProtocolError)  -> attempts = 3  (retried? True)
```

### Changelog

No `CHANGELOG.md` entry — [CONTRIBUTING.md](https://github.com/qBraid/qBraid/blob/main/CONTRIBUTING.md#submitting-a-pull-request) requires one under `[Unreleased]`. Two user-visible changes need covering: the new `QUANTINUUM_NEXUS_COMPILE_TIMEOUT` env var, and the retry behaviour.

### PR description understates the change

The Changes list covers only the timeout. Commit b77ed2c adds retry logic across four call sites — a substantive behavioural change that currently appears only in the CodeRabbit auto-summary. A reviewer reading the description wouldn't know to look at it.

### Not attributable to this PR

`black --check` fails on both files, but it fails identically on `main` — that's the repo-wide formatter drift, not something you introduced.

rryoung98:
Note on the check list: the push of 9f01bbcd (the two validation fixes) triggered only PR Compliance and CodeQL. The CI, Format and Docs workflows never fired for it, which looks like a missed webhook rather than anything about the change — all three ran normally on the previous commit ab4ad13e, and the PR is not a draft.

I re-ran the two that expose `workflow_dispatch` against the branch head, and both pass:

- CI (test matrix 3.10–3.14): https://github.com/qBraid/qBraid/actions/runs/30735586470 — success
- Docs: https://github.com/qBraid/qBraid/actions/runs/30735587510 — success

Those runs don't attach to this PR's check list because they were dispatched rather than triggered by `synchronize`, so the tests read as absent above even though they passed. `codecov/patch` did register from that run.

`format.yml` has no `workflow_dispatch` trigger, so it can't be re-run that way. Verified locally instead: `black --check` and `isort --check-only` are clean on all 7 changed files.

A close/reopen would fire `reopened` and re-attach everything properly. I've left that alone since the review request is open — happy to do it if you'd rather see the checks inline.

rryoung98:
Superseding my note above: the check list is healthy again. The push after merging `main` in triggered the webhook normally, so the full suite now runs inline on the head commit — test matrix 3.10–3.14, build, format, and `codecov/patch`, all passing. No need to follow the dispatched-run links or to close/reopen anything.

Two changes since your review, both in response to it:

- The transport work is no longer scoped to the compile wait. `qnexus` builds its process-wide `httpx.Client` with `timeout=None`, so bounding one call left the other five in `submit` able to pin a thread; there's now a client-level timeout applied at every qnexus entry point, and the retry covers `qnx_exc.ResourceCreateFailed`/`ResourceFetchFailed` gated on 502/503/504, since qnexus never raises `HTTPStatusError`. Backoff is halved and jittered for the reason you gave.
- Probing the job path while I was in there turned up three further defects, all fixed here: `status()` read `ref.last_status` off a cached ref and so could never observe a job finishing on the direct-SDK path; the status map covered 6 of `JobStatusEnum`'s 10 members, with `TERMINATED` and `DEPLETED` falling through to a non-terminal `UNKNOWN`; and `"E" in device_id.upper()` classifies `Helios-1` as a simulator.

On your point about self-reinforcing tests — that critique applied to more than the timeout test. `test_status_map_covers_expected_states` asserted the map equalled a hand-written copy of itself, and every status test built a `SimpleNamespace(last_status=...)`, which is the stale field the code was wrongly reading. Both are rewritten: the map is checked against `JobStatusEnum` itself, and the status tests drive a real `qnx.jobs.status`, one walking a ref that stays `SUBMITTED` while NEXUS progresses to `COMPLETED`. The counts fixtures were also palindromic (`{"00", "11"}`), so a bit-order flip was undetectable; they're asymmetric now.

Separately, live validation: 50 circuits through the managed path returned 44 passes and zero correctness failures. One of the failures was `httpx.RemoteProtocolError: Server disconnected without sending a response` during submit — the exact case the retry here handles, on the deployed build that lacks it.

rryoung98:
@ryanhill1 ready for another look. Everything from your review is addressed on the current head (`6e334c4`):

1. **Unbounded client.** `ensure_bounded_client()` in the new `_transport.py` sets a per-request timeout on qnexus's shared `httpx.Client` (`QUANTINUUM_NEXUS_HTTP_TIMEOUT`, default 60s). It runs at every qnexus entry point in the device and job classes, so all six calls in `submit` are bounded, not just `wait_for`.
2. **Retry missing 5xx.** Retries now also cover `ResourceCreateFailed` / `ResourceFetchFailed` gated on 502/503/504, so a NEXUS gateway failure is retried while a 4xx still surfaces immediately.
3. **Changelog.** Entries under `[Unreleased]` cover both env vars, the retry behaviour, and the job fixes.

The PR description was rewritten to cover the retry work and the qnexus 0.39.0 → 0.43 timeout-default regression you dug up.

Two notes. Fixing the blockers meant probing the job path, which turned up three further defects that are also in this PR (stale `status()` snapshot, 6/10 status map, `"E" in device_id` simulator detection); the description has the details, and tests went from 48 to 81. And yesterday's `ecfc2f6c` / `6e334c47` pair is a keepalive experiment and its revert, netting to zero; the revert message has the reasoning.

CI is green on 3.10 through 3.14; the only red check is `codecov/patch`.

pyproject.toml
```diff
@@ -53,7 +53,7 @@ oqc = ["oqc-qcaas-client>=3.11.0"]
 origin = ["pyqpanda3"]
 pasqal = ["pasqal-cloud>=0.20.3", "pulser-core>=1.4.0,<2.0"]
 qat = ["myqlm-interop>=2.0"]
-quantinuum = ["qnexus>=0.39.0", "pytket>=1.31"]
+quantinuum = ["qnexus>=0.39.0,!=0.48.0", "pytket>=1.31"]
```
ryanhill1: **B1 (blocking):** This excludes one version of a two-version bug. 0.48.1 has the identical undeclared-`selene_core` import and is the current latest release, so this constraint still resolves to a broken package:

```
$ pip install 'qnexus>=0.39.0,!=0.48.0' 'pytket>=1.31'
→ qnexus 0.48.1
→ ModuleNotFoundError: No module named 'selene_core'
```

Checked the wheels directly — `models/references/__init__.py` imports `selene_core` in 0.48.0 **and** 0.48.1, but not 0.46.0/0.47.0; none of the four declare it in `requires_dist`. Suggest `!=0.48.0,!=0.48.1` or `<0.48`.

Green CI doesn't contradict this: the full dev requirement set resolves qnexus back to **0.47.0**, so CI never installs 0.48.x. The published extra's two-package constraint set does. Worth a resolution check against the extra alone, since CI structurally can't catch this class of bug.

requirements-dev.txt
```diff
@@ -22,7 +22,7 @@ qiskit-qasm3-import>=0.5.1
 qiskit-aer>=0.15.0; python_version < "3.13"
 qiskit-ionq>=0.5.12
 pyqpanda3>=0.2.0; python_version < "3.13"
-qnexus>=0.39.0
+qnexus>=0.39.0,!=0.48.0
```
ryanhill1: Same as the pyproject change — see the note on `pyproject.toml:56`. This file is what pins CI to 0.47.0 today, which is why the gap is invisible here.

qbraid/runtime/quantinuum/job.py
```diff
@@ -47,7 +66,7 @@ class QuantinuumJobError(QbraidRuntimeError):
 
 
 def _map_quantinuum_status(last_status: str | None) -> JobStatus:
-    """Map a qnexus ``last_status`` string to a qBraid :class:`JobStatus`."""
+    """Map a qnexus job status value to a qBraid :class:`JobStatus`."""
     if last_status is None:
         return JobStatus.UNKNOWN
     return _QUANTINUUM_STATUS_MAP.get(last_status, JobStatus.UNKNOWN)
```
ryanhill1: **C1:** CONTRIBUTING.md L218-219 asks for direct indexing on vendor-keyed lookup tables specifically so an unmapped value raises rather than degrading, and L210-211 names this exact outcome — "a job whose status quietly reads `UNKNOWN` because the vendor added an enum member."

That's this PR's second defect. The map is now complete (verified 10/10 against `JobStatusEnum`), but the mechanism that let the gap go unnoticed is unchanged — an 11th NEXUS state reproduces it identically, silently. `qbraid/runtime/aqt/job.py:120-122` shows the compliant shape:

```python
try:
    return _STATUS_MAP[status]
except KeyError as err:
    raise AQTJobError(f"Unrecognized AQT job status '{status.value}'.") from err
```

Note this also inverts `test_status_unknown_on_unrecognized_value`, which currently pins the non-compliant behaviour. (Keying on strings rather than enum members is well justified in your comment and is independent of this.)

qbraid/runtime/quantinuum/device.py
```diff
@@ -122,39 +142,94 @@ def submit(  # pylint: disable=arguments-differ
...
+        try:
+            execute_job = qnx.start_execute_job(
+                programs=compiled_refs,
+                name=execute_name,
+                n_shots=[shots] * len(compiled_refs),
+                backend_config=backend_config,
+                project=project,
+                language=Language.QIR,
+            )
+        except (httpx.TransportError, ConnectionError) as err:
```
ryanhill1, line 234: **D1:** This catches the failure mode `_transport.py:25-27` argues is the *less* likely one. qnexus checks status codes by hand — `qnexus/client/jobs/_execute.py:169` raises `ResourceCreateFailed(status_code=...)` on any non-202 — so a gateway 502/503 on this dispatch never arrives as an `httpx.TransportError` and escapes bare, without the "check project X for job Y and cancel it" guidance you wrote for exactly this situation.

Since this is the one billable, deliberately-unretried call, suggest adding `qnx_exc.ResourceCreateFailed` to the except (still not retrying).

qbraid/runtime/quantinuum/_transport.py
```diff
@@ -0,0 +1,240 @@
...
+    try:
+        value = int(raw)
+    except ValueError as err:
+        raise QuantinuumDeviceError(f"{name} must be an integer, got {raw!r}.") from err
+    if not minimum <= value <= maximum:
+        raise QuantinuumDeviceError(f"{name} must be between {minimum} and {maximum}, got {value}.")
+    return value
+
+
+def ensure_bounded_client() -> None:
```
ryanhill1, line 118: **D2:** This is applied at every qnexus entry point in `device.py` and `job.py`, but not in `provider.py:35-40` (`_fetch_quantinuum_devices_df`). In a fresh process, device enumeration is typically the *first* qnexus call a server makes, and it runs unbounded:

```
client.timeout AFTER provider.get_devices():   Timeout(timeout=None)
client.timeout AFTER ensure_bounded_client():  Timeout(timeout=60.0)
```

Because this mutates a process-wide singleton, any later device/job call retroactively fixes it — but the first listing is exposed, which is the `qbraid-runtime-api` startup path. The CHANGELOG's "Every NEXUS request now carries a timeout" isn't true until this covers the provider.

qbraid/runtime/quantinuum/job.py
```diff
@@ -172,13 +213,15 @@ def result(self) -> Result[GateModelResultData]:
...
+            detail = self._status_detail
+            message = (
+                (detail.error_detail or detail.message) if detail is not None else None
+            ) or "no error message"
+            raise QuantinuumJobError(f"Quantinuum job {self.id} failed: {message}")
 
         ref = self._get_ref()
         try:
+            ensure_bounded_client()
             results = qnx.jobs.results(ref)
```
ryanhill1, line 248: **D3:** `download_result()` at line 239 is lazy in exactly the way you identified for `CompilationResultRef.get_output` — it calls `_get_execute_results()` and issues its own HTTP request — but the loop sits outside this `try`, so a blip there escapes as a raw qnexus/httpx error instead of `QuantinuumJobError`, unlike every other qnexus call in the file. Pre-existing line, but the same insight this PR applies in `device.py:207-209`. Extending the `try` to cover the loop would close it.

qbraid/runtime/quantinuum/job.py
```diff
@@ -146,14 +188,13 @@ def execution_time_s(self) -> float | None:
         """
         if self.status() != JobStatus.COMPLETED:
             return None
-        ref = self._get_ref()
-        last_status_detail = ref.last_status_detail
-        if last_status_detail is None:
+        detail = self._status_detail
```
ryanhill1: **R1:** `status()` early-returns on a cached terminal status without populating `_status_detail`, so this raises "status detail is missing" for a genuinely completed job. Base `QuantumJob.__init__` does `self._cache_metadata = {"job_id": job_id, **kwargs}`, so any caller passing `status=` triggers it.

Not reachable from the SDK path today, but there's precedent in the sibling repo — `qbraid-runtime-api/app/runtime/ionq/job_actions.py:51` does `job._cache_metadata.update(job_data)`. Worth confirming against #261 before it becomes reachable; a fetch-if-`None` fallback here would make it moot.

qbraid/runtime/quantinuum/device.py
```diff
@@ -122,39 +142,94 @@ def submit(  # pylint: disable=arguments-differ
...
         def unique(label: str) -> str:
             ts = datetime.now(timezone.utc).strftime("%Y%m%dT%H%M%S")
             return f"qbraid {label} {ts}-{uuid.uuid4().hex[:6]}"
 
+        # Pre-execute stages are retried on transient connection errors; a repeat
+        # at worst orphans an upload or compile job, with no execution cost.
         circuit_refs = [
-            qnx.circuits.upload(name=unique(f"circuit-{i}"), circuit=c, project=project)
+            retry_transient(
+                lambda i=i, c=c: qnx.circuits.upload(
```
ryanhill1: **C3:** This lambda's default-argument binding introduces the one new mypy error in the module — `device.py:167: Cannot infer type of lambda [misc]` (main: 13 errors, this branch: 14). CONTRIBUTING L156 lists mypy as enforced, though in practice it's absent from tox `envlist`, from `format-check`, and from every CI workflow — so nothing could have flagged it. A named inner function or `functools.partial` avoids it. `_transport.py` itself is mypy-clean.

ryanhill1 (commented):
Thanks for this — it's a genuinely thorough pass, and the two job-path defects are good catches that a diff-level review would never have surfaced. Removing the tautological `test_status_map_covers_expected_states` is the kind of fix that pays off repeatedly.

**What I verified independently** (worktree + live Nexus session): the status map now covers all 10 `JobStatusEnum` members exactly, with no extras; `INITIALIZING` is indeed absent from the enum; qnexus does build one process-wide `httpx.Client(timeout=None)` (`client/__init__.py:183`); `wait_for`'s default is now `None` and it raises `asyncio.TimeoutError`, which your handler catches correctly on 3.10 through 3.14; `_import_or_skip` behaves as documented. pylint 10.00/10, black/isort/ruff clean, 86 tests pass. Broader `tests/runtime`: 809 passed, 5 failed — all 5 fail identically on `main` (missing `cirq`/`icalendar`, azure/numpy), so no regressions from this PR.

I also checked the fixture-quality guidance and it holds up — the status tests build real `NexusJobStatus` objects rather than namespaces, and the one `SimpleNamespace` is unavoidable since an invalid `JobStatusEnum` can't be constructed.

Two things I'd hold on: **B1** (`pyproject.toml:56`), because the extra still installs broken today, and **C1** (`job.py:72`), because the PR's own bug class stays reachable. **D1/D2/D3** are real but arguable in scope. The rest are nits, marked as such.

One process gap with no line to anchor to: **C2** — there's no remote test for this integration. CONTRIBUTING L126-131 asks for verification against the live API with the credentialed tests kept behind the `remote` marker, and twelve other vendors have a `test_*_remote.py` (aqt, aws, azure, ionq, oqc, openquantum, origin, pasqal, …). Quantinuum has none. Cheap to add — `qnx.devices.get_all()` returned 4 devices in seconds while I was reviewing this, and that payload is exactly the fixture basis the guideline describes.

Minor FYI, no action needed: qnexus has a second unbounded client, `_get_auth_client()` (`auth.py:66`, `timeout=None`), which `ensure_bounded_client()` doesn't reach. It's login-only — the per-request 401 refresh goes through `auth_flow`'s `yield` and does inherit the bounded client — so the server path is fine. Only relevant if anything ever calls `login_with_credentials()` at startup.

qbraid/runtime/quantinuum/_transport.py
```diff
@@ -0,0 +1,240 @@
...
+
+#: Gateway-class responses that are worth repeating. Anything else (400s, and
+#: 500s that indicate a rejected program rather than an unavailable upstream)
+#: is a real failure and is raised immediately.
+RETRYABLE_STATUS_CODES = frozenset({502, 503, 504})
+
+_HTTP_TIMEOUT_ENV = "QUANTINUUM_NEXUS_HTTP_TIMEOUT"
+
+
+class QuantinuumDeviceError(QbraidRuntimeError):
```
ryanhill1: **C5 (nit):** Moving `QuantinuumDeviceError` here makes its `__module__` resolve to `qbraid.runtime.quantinuum._transport` — the path users see in tracebacks and Sphinx — and `_transport.py` is the only private module in any provider package. It's also not exported from `qbraid.runtime.quantinuum` (that import raises), while the CHANGELOG now points users at it.

I assume the circular import drove this. A non-underscore `exceptions.py` would avoid both. Export convention is genuinely mixed here (aqt and ionq export theirs; azure/oqc/pasqal/origin/rigetti don't), so take or leave — your call.

CHANGELOG.md
```diff
@@ -66,6 +72,7 @@ Writing an entry:
 - Fixed `QuantinuumDevice.status()` raising `ResourceFetchFailed` (400 "Invalid machine name") for cloud-hosted NEXUS emulators such as `H2-Emulator`, which broke `device.run()` since pre-submit validation checks status. Devices now carry a `nexus_hosted` profile flag, and cloud-hosted emulators report `ONLINE` without calling the hardware-only machine status endpoint ([#1295](https://github.com/qBraid/qBraid/pull/1295))
 
 ### Dependencies
+- Excluded `qnexus` 0.48.0 from the `quantinuum` extra (`>=0.39.0,!=0.48.0`). That release imports `selene_core` without declaring it as a dependency, so `import qnexus` raises `ModuleNotFoundError` on a clean install. In CI the broken import made `pytest.importorskip` silently skip the entire Quantinuum test file — every job green, patch coverage 0% — so the tests now skip only when qnexus itself is absent and fail loudly when it is installed but broken ([#1297](https://github.com/qBraid/qBraid/pull/1297))
```
ryanhill1: **B1 follow-on:** this entry will need rewording once the constraint covers 0.48.1 — as written it names only 0.48.0.

**C4 (nit):** At 70 words / 3 sentences this is over the file's own budget ("one or two sentences, roughly 40 words"), and the middle clause — "In CI the broken import made `pytest.importorskip` silently skip the entire Quantinuum test file — every job green, patch coverage 0%" — is the diagnosis the style note asks to keep in the PR description. Consistent either way; the surrounding entries are longer still, and the note anticipates that ratchet.

CHANGELOG.md
```diff
@@ -48,6 +48,12 @@ Writing an entry:
...
 - Fixed `openqasm3_to_pyquil` padding declared-but-idle qubits above the highest qubit in use with identity, making a program claim more of the QPU than it has work for ([#1301](https://github.com/qBraid/qBraid/pull/1301))
 - Fixed `pyquil_to_cirq` rejecting `RXX`/`RYY`/`RZZ` as "noise gates". They now convert to Cirq's `XXPowGate`/`YYPowGate`/`ZZPowGate` ([#1301](https://github.com/qBraid/qBraid/pull/1301))
+- Fixed `QuantinuumJob.status()` never observing a job finish. It read a snapshot field on the job reference that is fixed when the reference is built, so `job = device.run(...)` followed by `job.result()` polled a submission-time status until it timed out. Status is now fetched from NEXUS on each call ([#1297](https://github.com/qBraid/qBraid/pull/1297))
+- Fixed the Quantinuum job status map covering 6 of NEXUS's 10 states. `TERMINATED` and `DEPLETED` (credits exhausted) are terminal on NEXUS but mapped to the non-terminal `UNKNOWN`, so a stopped job was polled indefinitely; `SUBMITTED`, `CANCELLING` and `RETRYING` were also unmapped ([#1297](https://github.com/qBraid/qBraid/pull/1297))
```
ryanhill1: **N1 (nit):** "covering 6 of NEXUS's 10 states" — the old map had 6 *keys*, but `INITIALIZING` isn't a `JobStatusEnum` member at all, so it covered 5. The entry slightly understates the bug.

pyproject.toml
rryoung98: Went with `<0.48`: 0.48.2 released this morning and still imports `selene_core` without declaring it (checked the wheel), so per-version exclusions are a losing race against their release cadence. Both this file and requirements-dev.txt updated in 667fa147.

requirements-dev.txt
rryoung98: Updated together with pyproject.toml in 667fa147 (`<0.48`).

qbraid/runtime/quantinuum/job.py
rryoung98: Done in 667fa147: the map is now indexed directly and an unmapped status raises `QuantinuumJobError` naming the value, matching the AQT shape. `test_status_unknown_on_unrecognized_value` is inverted to `test_status_raises_on_unrecognized_value`, and there is a unit test on the mapper itself.

qbraid/runtime/quantinuum/device.py
rryoung98: Done in 667fa147. `ResourceCreateFailed` with a 502/503/504 now gets the same check-project-X-for-job-Y guidance (a gateway can fail after forwarding the request, so acceptance is just as ambiguous as a timeout); any other status code propagates untouched, since a 4xx means NEXUS refused the job and the orphan wording would mislead. Still never retried. Two new tests cover both branches.

qbraid/runtime/quantinuum/_transport.py
rryoung98: Done in 667fa147: `_fetch_quantinuum_devices_df` now calls `ensure_bounded_client()` itself, so the first call a fresh server process makes is bounded. Test added.

qbraid/runtime/quantinuum/job.py
rryoung98: Done in 667fa147: the `download_result()` loop is wrapped and surfaces as `QuantinuumJobError("Failed to download results...")`. Test added.

qbraid/runtime/quantinuum/job.py
rryoung98: Done in 667fa147: new `_get_status_detail()` fetches on demand when the cached-terminal early return left `_status_detail` unset, so a job constructed with `status=COMPLETED` computes timing instead of raising. Covered by two new tests, including the wrapped-error path.

qbraid/runtime/quantinuum/device.py
rryoung98: Done in 667fa147: the default-arg lambda is now `functools.partial` over a named `upload_circuit` closure. `mypy qbraid/runtime/quantinuum/` is back to the 13-error module baseline, with no lambda-inference error.

qbraid/runtime/quantinuum/_transport.py
rryoung98: Took it in 667fa147: both `QuantinuumDeviceError` and `QuantinuumJobError` now live in a public `exceptions.py` and are exported from `qbraid.runtime.quantinuum` (docs autosummary included). The old `.device`/`.job`/`._transport` import paths still work.

CHANGELOG.md
rryoung98: Reworded in 667fa147: the entry now covers the whole 0.48.x range and drops the CI-diagnosis clause (that story stays in the PR description).

CHANGELOG.md
rryoung98: Corrected to 5 in 667fa147, with a parenthetical noting `INITIALIZING` was never a NEXUS status.

rryoung98:
@ryanhill1 all ten findings from the second round are addressed on `667fa147`, with a threaded reply on each. The one judgment call: the qnexus constraint is `<0.48` rather than per-version exclusions, because 0.48.2 shipped this morning with the same undeclared `selene_core` import — third broken release in the series. I also took the C5 nit: both exceptions now live in a public `exceptions.py` and are exported from the package, with the old import paths intact.

Quantinuum suite is 91 passing (was 81); black/isort/pylint clean on the changed files; `mypy qbraid/runtime/quantinuum/` is back to the 13-error baseline.

ryanhill1 (approved):

PR #1312, opened 2026-08-01
fix(transpiler): select conversion paths by weight, not hop count

### LINKED ISSUE #1307: transpile() selects conversion paths by hop count, ignoring declared conversion weights
## Summary

`transpile()` selects a conversion path by **hop count only**. Conversion weights are declared, stored on the graph edges, and honored by one path-finding method, but the method `transpile()` actually calls ignores them entirely. Five conversions carry deliberately reduced weights that have never influenced routing.

This is not theoretical: the weight system already encoded the fix for a bug that reached production.

## The two APIs disagree

`ConversionGraph` has two path finders.

`find_shortest_conversion_path` honors weights (`qbraid/transpiler/graph.py:321`):

```python
path = rx.dijkstra_shortest_paths(
    self,
    self._node_alias_id_map[source],
    target=self._node_alias_id_map[target],
    weight_fn=lambda edge: edge["weight"],
)
```

`find_top_shortest_conversion_paths` does not (`graph.py:363`):

```python
sorted_paths = sorted(all_paths, key=len)[:top_n]
```

`transpile()` calls the second one (`qbraid/transpiler/converter.py:112`):

```python
paths = graph.find_top_shortest_conversion_paths(source, target, top_n=max_path_attempts)
```

So for every user-facing conversion, `edge["weight"]` is written but never read. Ties among equal-length paths break on edge insertion order.

## The weights are meaningful, and non-default

32 conversions declare a weight. Six are below 1.0:

| conversion | weight |
| --- | --- |
| `cirq_to_pyquil` | **0.74** |
| `cirq_to_braket` | 0.85 |
| `openqasm3_to_cudaq` | 0.95 |
| `braket_to_cirq` | 0.99 |
| `qiskit_to_qasm2` | 0.999 |

These look like deliberate quality rankings, and `ConversionGraph.__init__` documents an `edge_bias` mechanism built on them (`graph.py:117-120`):

> `edge_bias (float, optional): Factor used to fine-tune the edge weight calculations ... weight 0.78 over two conversions at weight 1.0. Only used if conversions is None.`

None of that reaches `transpile()`.

## Why this matters: it already caused a production bug

`cirq_to_pyquil` carries the lowest weight in the codebase, 0.74. Someone had already identified it as the least trustworthy conversion available.

`transpile(qasm2_program, "pyquil")` nevertheless routed through it, because `qasm2 -> cirq -> pyquil` is 2 hops while the correct `qasm2 -> qasm3 -> openqasm3 -> pyquil` is 3. Going through Cirq reorders instructions across measurement boundaries and fragments the readout register into one `BIT[1]` per measurement key, which produced `Misplaced or illegal instruction in ProtoQuil program` on 23 production Rigetti jobs.

A weighted selection would have preferred the 3-hop path and avoided it. PR #1301 fixes the symptom by adding a direct `qasm2 -> pyquil` converter, which wins on hop count. The underlying selection issue is untouched.

## Suggested direction

Not obvious which way to go, so flagging rather than prescribing:

1. **Make `transpile()` weight-aware.** Most faithful to the declared design, but it changes routing for existing conversions and needs a careful audit of what moves. `find_top_shortest_conversion_paths` returns N candidates for retry, so it needs a weighted top-N (e.g. sort by summed weight, or Yen's algorithm), not just Dijkstra.
2. **Have the two methods agree**, whichever semantics win, so `shortest_path()` stops reporting a path `transpile()` will not take. The divergence itself is a debugging hazard: it made the root cause of #1301 hard to establish, because the two APIs give different answers for the same pair.
3. **If weights are genuinely not wanted for routing**, remove them and the `edge_bias` documentation, so nobody sets a weight expecting an effect.

At minimum, the current state should not persist silently: a contributor adding `@weight(0.5)` today gets no error and no effect.

## Notes

- Found while root-causing #1301. Not urgent, since hop count happens to give acceptable answers today, but it is a latent trap for any future conversion added on the assumption that weights work.
- Tie-breaking on insertion order is a related sharp edge: adding an unrelated edge can silently reroute an existing pair.

ryanhill1:
Closes #1307.

`find_top_shortest_conversion_paths` ranked candidates with `sorted(all_paths, key=len)`, and that is the method `transpile()` calls. So `edge["weight"]` was written on every edge and read by nothing on the user-facing path.

## Change

Rank paths by summed conversion cost — `log(1/weight) + bias`, the quantity the *other* path finder was already minimizing. Ties fall back to hop count, then node aliases, so adding an unrelated conversion can no longer silently reroute an existing pair. `find_shortest_conversion_path` ranks by the same key (direction 2 from the issue), reaching it by Dijkstra rather than enumeration.

## Blast radius: 38 of 213 reachable pairs

Diffing the route `transpile()` actually takes, `main@4cdb5fdc` vs this branch.

**3 hop-count reroutes**, all avoiding `cirq_to_pyquil` (weight 0.74, lowest in the codebase):

```
cirq   -> pyquil   cirq -> pyquil            (0.551)  ->  cirq -> qasm2 -> pyquil            (0.500)
braket -> pyquil   braket -> cirq -> pyquil  (0.811)  ->  braket -> cirq -> qasm2 -> pyquil  (0.750)
stim   -> pyquil   stim -> cirq -> pyquil    (0.801)  ->  stim -> cirq -> qasm2 -> pyquil    (0.750)
```

**35 equal-hop tie-breaks** — exact cost ties under the shipped weights, now resolved by node alias instead of by rustworkx enumeration order (e.g. `qasm3 -> qiskit -> qasm2` becomes `qasm3 -> cirq -> qasm2`).

## Why the reroute is right, not just declared

The issue's caution was that 0.74 might predate `openqasm3_to_pyquil` and #1301. Checked on a Bell circuit with two measurements:

```
cirq -> pyquil                     cirq -> qasm2 -> pyquil
DECLARE m0 BIT[1]                  DECLARE ro BIT[2]
DECLARE m1 BIT[1]                  H 0
H 0                                CNOT 0 1
CNOT 0 1                           MEASURE 0 ro[0]
MEASURE 0 m0[0]                    MEASURE 1 ro[1]
MEASURE 1 m1[0]
```

That fragmented readout is the shape that failed 23 production Rigetti jobs with `Misplaced or illegal instruction in ProtoQuil program`. The weight was right; hop-count ranking took the edge anyway because it is a single hop.

## Weights changed

`qiskit_to_qasm2` 0.999 → **1.0** (came from an unrelated 2025 PR, matched no benchmark, existed only as a hidden tie-break — it is what makes six of the 35 tie-breaks exact) and `braket_to_cirq` 0.99 → **1.0**, same reasoning. `cirq_to_pyquil` stays at 0.74, rationale now in a comment: ~0.88 gate fidelity direct vs ~0.95 through `qasm2`, so it should keep losing multi-hop comparisons even though its original justification is now fixed.

## Measurement handling

Coupled to the reroute rationale, so it ships here: `cirq_to_braket`, `pytket_to_braket`, `braket_to_cirq`, and `cirq_to_pyquil` no longer drop or fragment measurements, and `cudaq_to_qasm2` raises instead of passing `"{translation failed}"` off as success. `test_measurement_coverage.py` sweeps every reachable pair with a measured GHZ circuit.

## Bugs found reviewing the above

Each has a regression test:

- **Merged readout bits were ordered by moment, not measurement key** — a circuit measuring q1 before q0, or QASM with a permuted `creg` like `measure q[2] -> c[0]`, came out transposed. Silently, which is the failure mode merging exists to prevent.
- **Path finding enumerated every simple path** — factorial in density: ~3.3s per call on a dense 11-node graph, and at 14 nodes it did not finish in 10 minutes. The native graph has three simple paths for `qasm2 -> pyquil` so CI saw nothing, but `ConversionGraph` takes arbitrary user conversions. The single-path search is now Dijkstra and top-N is Yen's algorithm; a dense 18-node graph answers `top_n=3` in 0.4 ms. Verified indistinguishable from enumerate-and-sort across 3,408 (pair, k, depth) combinations on the native graph and 5,460 on generated graphs.
- **`max_path_depth` was applied after truncating to `max_path_attempts`** — a depth limit reported "no path" whenever the top few candidates all exceeded it, even when a qualifying path ranked just below them. It is now applied during selection.
- **`source == target` returned a round trip** (`qasm2 -> cirq -> qasm2`) where main raised `ConversionPathNotFoundError` — `all_simple_paths` returns cycles.
- Two smaller ones in the merge: the fixed `"m"` key collided with a mid-circuit measurement already using it, producing a circuit cirq rejects on simulation; and a `confusion_map` on a terminal measurement was dropped silently, where `cirq_to_braket` raises for the same construct.

## Caveats

`tests/transpiler` + `tests/programs` + `tests/runtime/rigetti`: 1731 passed, 8 `cudaq` failures identical with and without these changes (an earlier revision of this description reported pyqir/qiskit_qir failures — different local environment). Lint clean.

The blast radius was re-measured after Yen's replaced the enumeration and is byte-identical: still 38 of 213 pairs, 3 reroutes and 35 tie-breaks.

**Also in this PR:**
- Updates Cirq, Braket, PyQuil, CUDA-Q, pytket, and Qiskit conversion handling alongside the routing changes.
- Adds measurement-coverage and measurement-semantics regression tests across supported program formats.
- Adds CUDA-Q-to-QASM2 conversion coverage and expands conversion edge and graph test coverage.
- Updates the changelog and program-equality test expectations for the revised conversion behavior.

<details>
<summary>Conversion Routing and Execution</summary>

```mermaid
sequenceDiagram
  participant Converter as converter.py
  participant Graph as graph.py ⚠️
  participant Edge as edge.py ⚠️
  participant CirqPyQuil as cirq_to_pyquil.py ⚠️
  participant CirqBraket as cirq_to_braket.py
  participant BraketCirq as braket_to_cirq.py
  participant CudaQ as cudaq_to_qasm2.py ⚠️
  participant Pytket as pytket_extras.py ⚠️
  Converter->>Graph: find conversion path
  Graph->>Edge: evaluate weighted edges
  Edge-->>Graph: path cost
  Graph-->>Converter: selected conversion path
  Converter->>CirqPyQuil: execute direct conversion
  Converter->>CirqBraket: execute Cirq to Braket conversion
  Converter->>BraketCirq: execute Braket to Cirq conversion
  Converter->>CudaQ: execute CUDA-Q to QASM2 conversion
  Converter->>Pytket: convert pytket classical bits
```
</details>

<details>
<summary>Changed Transpiler Module Dependencies</summary>

```mermaid
graph LR
  Converter[converter.py] --> Graph[graph.py ⚠️]
  Graph --> Edge[edge.py ⚠️]
  Converter --> CirqPyQuil[cirq_to_pyquil.py ⚠️]
  Converter --> CirqBraket[cirq_to_braket.py]
  Converter --> BraketCirq[braket_to_cirq.py]
  Converter --> CudaQ[cudaq_to_qasm2.py ⚠️]
  Converter --> Pytket[pytket_extras.py ⚠️]
  Converter --> Qiskit[qiskit_to_qasm2.py]
  CirqBraket --> BraketCirq
```
</details>

<sub>Auto-enriched by [Argus](https://argus.reviews)</sub>

## Summary by CodeRabbit

* **New Features**
  * Improved conversion-path selection with weighted ranking, deterministic tie-breaking, and optional depth limits.
  * Preserved measurement ordering, classical-bit mappings, and inversion behavior across supported framework conversions.
  * Added clearer errors for unsupported mid-circuit measurements, reused classical bits, and failed CUDA-Q translations.

* **Bug Fixes**
  * Corrected conversion weights and improved handling of zero-weight routes.
  * Fixed measurement loss and ordering issues during circuit conversion.

* **Documentation**
  * Updated the unreleased changelog with conversion, routing, measurement, and error-handling improvements.


tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -170,14 +179,17 @@ def test_qrisp_coverage(target, baseline, qrisp_circuits, conversion_graph):
...
     for gate_name in qrisp_circuits:
+        if gate_name in skips:
+            continue
         try:
             convert_from_qrisp_to_x(target, gate_name, qrisp_circuits, conversion_graph)
         except Exception as e:  # pylint: disable=broad-exception-caught
             failures[f"{target}-{gate_name}"] = e
 
-    total_tests = len(qrisp_circuits)
+    total_tests = len(qrisp_circuits) - len(skips)
```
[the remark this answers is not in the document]
ryanhill1: Applied in e562b1e5 — the denominator now counts only attempted gates, so a skip entry absent from `qrisp_circuits` can't distort the ratio.

tests/transpiler/test_measurement_coverage.py
```diff
@@ -0,0 +1,166 @@
...
+            f"{source} -> {target} now preserves measurements; "
+            "remove it from KNOWN_MEASUREMENT_FAILURES"
+        )
+        return
+
+    assert (
+        n_measure == NUM_QUBITS
+    ), f"{source} -> {target}: expected {NUM_QUBITS} measurements, found {n_measure}"
+    if target == "pyquil":
+        assert_pyquil_readout_intact(out)
```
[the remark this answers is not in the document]
ryanhill1: Valid concern, addressed with a dedicated direct-call test rather than a special case in this sweep: `test_multi_key_terminal_measurements_merge_to_single_register` (tests/transpiler/pyquil/test_conversions_pyquil.py, added in 3c0550a9) calls `cirq_to_pyquil` directly with per-bit measurement keys and asserts the single `DECLARE m0 BIT[3]` plus per-index MEASURE targets — precisely the `_merge_terminal_measurements` path `transpile()` routes around. Keeping this sweep converter-agnostic.

qbraid/transpiler/graph.py
```diff
@@ -303,10 +304,137 @@ def remove_conversion(self, source: str, target: str) -> None:
...
+        target_id: int,
+        banned_nodes: frozenset[int] = frozenset(),
+        banned_edges: frozenset[tuple[int, int]] = frozenset(),
+    ) -> list[int] | None:
+        """Return the best-ranked path between two node ids, or None if unreachable.
+
+        Explores in ranking order, so the first time the target is reached its path is the
+        one :meth:`find_top_shortest_conversion_paths` ranks first. Appending an edge can
+        only increase the key -- cost is non-negative and the hop count strictly grows -- so
+        the greedy order is exact. Unlike enumerating every simple path, this stays
```
ryanhill1, line 342: **D1 (defect):** "the greedy order is exact" — not when a weight-0 edge is present. `inf + x = inf` collapses cost distinctions, so path order is *not* preserved under extension (the key is monotone but not isotone), and Dijkstra's settled-node discard throws away the true winner:

```python
specs = [("a","x",1.0), ("x","m",1.0), ("a","m",0.0), ("m","t",0.0)]
g = ConversionGraph(conversions=[Conversion(s,t,lambda x:x,w,bias=0.25) for s,t,w in specs])
g.shortest_path("a", "t")   # -> 'a -> x -> m -> t'
# enumerate-and-sort with _rank_key says: 'a -> m -> t'  (inf ties, 2 hops < 3 hops)
```

`m` settles via the finite `a->x->m` (beats `a->m`'s inf), then the shared inf edge `m->t` equalizes the costs — at which point `a->m->t` should win on hops, but it was already discarded. Root-seeding (see D2) fixes Yen's decomposition but not this settling problem; the clean fix is a large **finite** cost for weight 0 in `edge.py`, which restores translation-invariance for the whole key. Fuzz across 30 seeded graphs with zero-weight edges: 219 top-N mismatches vs the enumerate-and-sort definition on this implementation; `test_ranking_holds_with_infinite_cost_edges` passes only because its graphs are too small to trigger the settling.

qbraid/transpiler/graph.py
```diff
@@ -303,10 +304,137 @@ def remove_conversion(self, source: str, target: str) -> None:
...
+        self, source_id: int, target_id: int, k: int, max_depth: int | None = None
+    ) -> Iterator[list[int]]:
+        """Yield up to ``k`` paths as node ids, cheapest first, by Yen's algorithm.
+
+        Each successive path is the best one that deviates from an already-accepted path at
+        some node, found by re-running :meth:`_cheapest_path` with the root banned. That costs
+        ``O(k * V * (E + V log V))``, where enumerating every simple path and sorting is
+        factorial in graph density.
+
+        Yen's usual proof assumes additive scalar costs; it carries over to the composite
```
ryanhill1, line 403: **D2 (defect):** the proof carry-over argument has a float-shaped hole even with all-finite weights. `_rank_key` sums edge costs left-to-right from the *source*; the spur Dijkstra accumulates from the *spur node*. For two mathematically tied deviations (same multiset of edge costs — common with duplicate-heavy weights like this graph's 1.0s), the two groupings round differently in the last ulp, so the spur search can pick the deviation that the full-path ranking orders *second*. The losing candidate is then absent from the pool at selection time and paths are emitted out of order — first seen at `top_n=4` in my fuzz, i.e. within `transpile()`'s reach on a user graph. 19/12510 comparisons diverged from enumerate-and-sort on random graphs with weight pool {1.0, 0.9, 0.74, 0.5, 0.25} (no zeros).

Verified fix: pass the root's accumulated `(cost, hops, aliases)` into `_cheapest_path` as the heap's initial key. All accumulation becomes the same left fold `sum()` performs, spur-optimal becomes total-optimal by construction, and the same fuzz goes to 0/16680. It also simplifies the docstring: the spur search then literally minimizes the ranking key of the whole candidate.

qbraid/transpiler/graph.py
```diff
@@ -303,10 +304,137 @@ def remove_conversion(self, source: str, target: str) -> None:
...
+            return
+
+        alias_of = self._alias_of()
+        accepted, candidates = [best], []
+        yielded = 0
+
+        def qualifies(path: list[int]) -> bool:
+            return max_depth is None or len(path) - 1 <= max_depth
+
+        if qualifies(best):
```
ryanhill1: **N5 (nit):** the first `yield` runs before any `yielded < k` comparison, so `top_n=0` (or negative) returns one path where main's `sorted(...)[:0]` produced none and raised `ConversionPathNotFoundError`. Degenerate input, but it's a silent behavior change on public API — `if k <= 0: return` before yielding `best` restores it (or validate `top_n >= 1` in the caller).

qbraid/transpiler/conversions/braket/braket_to_cirq.py
```diff
@@ -86,16 +86,25 @@ def braket_to_cirq(circuit: BKCircuit) -> cirq_circuits.Circuit:
...
-    circuit = cirq.Circuit(
-        _from_braket_instruction(instr, qubit_mapping) for instr in circuit.instructions
+    # Braket measurements are always terminal and carry no key, so collect them into a
+    # single keyed measurement; per-instruction M('') gates fragment or collapse the
+    # readout register in downstream QASM/Quil exports.
+    measured_qubits = []
+    gate_instructions = []
+    for instr in circuit.instructions:
+        if str(instr.operator) == "Measure":
+            measured_qubits.extend(qubit_mapping[int(q)] for q in instr.target)
```
ryanhill1: **D3 (defect):** merged readout order follows *instruction* order, but Braket's readout order is the `Measure` operator's classical bit index (`_target_index`), and the two can disagree — `Circuit.from_ir` preserves permuted mappings:

```python
qasm = 'OPENQASM 3.0; bit[3] b; qubit[3] q; x q[0]; x q[1]; b[2] = measure q[0]; b[1] = measure q[1]; b[0] = measure q[2];'
bk = Circuit.from_ir(OpenQasmProgram(source=qasm))
LocalSimulator("braket_sv").run(bk, shots=3).result().measurement_counts  # {'011': 3}
cirq.Simulator().run(braket_to_cirq(bk), repetitions=1).measurements     # {'m': [1, 1, 0]}  <- transposed
```

This is the same bug class as the moment-order merge this PR fixes in `cirq_to_pyquil` — order by the declared bit, not by encounter order. Collect `(instr.operator._target_index, qubit)` and sort by index (falling back to encounter order when indices are unset/duplicated); `test_braket_to_cirq_preserves_readout_order` doesn't catch it because it measures in ascending order, where the two orders coincide. Braket 1.120 exposes no public accessor for the index, so `getattr(instr.operator, "_target_index", None)` with a fallback is about the best available.

qbraid/transpiler/conversions/pytket/pytket_extras.py
```diff
@@ -43,13 +44,31 @@
...
     """
-    braket_circuit, _, _ = pytket_braket.braket_convert.tk_to_braket(circuit)
+    from pytket.circuit import OpType  # pylint: disable=import-outside-toplevel
+
+    braket_circuit, _, measure_map = pytket_braket.braket_convert.tk_to_braket(circuit)
+
+    measured_qubits = set()
+    for command in circuit.get_commands():
+        if command.op.type == OpType.Measure:
+            measured_qubits.update(command.qubits)
```
ryanhill1: **D4 (defect):** a second `Measure` on an already-measured qubit takes this branch (it's a `Measure`, so the `elif` guard never sees it), and since `measure_map` is keyed by qubit, one of the two classical bits silently disappears:

```python
tk = TKCircuit(1, 2); tk.X(0); tk.Measure(0, 0); tk.Measure(0, 1)
pytket_to_braket(tk)   # converts "successfully" with ONE measure — a declared bit is gone
```

Re-measuring a qubit is mid-circuit measurement by another name (Braket can't represent it), so the check that already exists one line down should also fire here: raise when `command.qubits` intersects `measured_qubits` *before* updating the set.

qbraid/transpiler/conversions/cirq/cirq_to_pyquil.py
```diff
@@ -39,6 +41,95 @@
...
+    measurement on q_0 may appear earlier. Keys without an index fall back to qubit order,
+    matching the readout convention the Braket converters document.
+    """
+    indexed = []
+    for op in terminal:
+        match = _BIT_INDEX.match(op.gate.key)
+        if match is None or len(op.qubits) != 1:
+            return sorted(terminal, key=lambda op: min(op.qubits))
+        indexed.append(((match["register"], int(match["index"])), op))
+    return [op for _, op in sorted(indexed, key=lambda pair: pair[0])]
```
ryanhill1, line 82: **N7 (nit):** sorting on `(register, index)` orders multiple registers alphabetically, but the `qasm2 -> pyquil` edge (and the #1314 Rigetti ruling) use *declaration* order — so the direct edge and the routed path disagree on multi-`creg` circuits:

```
# creg d[1]; creg c[1]; measure q[0]->d[0]; measure q[1]->c[0];
direct cirq->pyquil:  MEASURE 1 m0[0]; MEASURE 0 m0[1]   (c first, alphabetical)
qasm2->pyquil route:  MEASURE 0 ro[0]; MEASURE 1 ro[1]   (d first, DECLARE order)
```

Declaration order is unrecoverable from cirq keys, so this is likely unfixable here — but since `transpile()` normally routes around this edge anyway, a sentence in the docstring acknowledging the divergence would spare the next person the archaeology. Take or leave.

ryanhill1 (commented):
Adversarial review, verified by reproduction. Environment: Python 3.12, cirq 1.6.1, braket 1.120.0, pytket 2.18.1, qiskit 2.3.0, pyquil 5.0.0rc3.

**What passed:**
- `tests/transpiler` + `tests/programs`: **1449 passed, 58 skipped**, no failures.
- black, isort, ruff, pylint (10.00/10) all clean on the changed files.
- Blast radius re-measured independently on current `main` (including the AQT merge): **37 of 158 reachable pairs changed in this env — 2 reroutes + 35 tie-breaks** — consistent with the claimed 38/213 (the third reroute, `stim -> pyquil`, needs `stimcirq`, absent here; the 35 tie-breaks match exactly).
- Yen's = enumerate-and-sort on the native graph, and on my own fuzz for all finite-weight, non-near-tie cases.
- The measurement-semantics tests genuinely execute circuits, and the converter fixes they pin (Braket bit indices, pytket bit order, key-order merge) all hold up under my probes.

**Findings** (inline comments carry the repros):

| ID | Severity | Summary |
|---|---|---|
| D1 | defect | Weight-0 edges (cost `inf`) make **both** finders inexact — even `shortest_path()` returns a provably wrong path. `inf + x = inf` destroys the translation-invariance the Dijkstra/Yen's proofs need. |
| D2 | defect | Float accumulation order breaks Yen's near cost ties: 19/12510 mismatches vs enumerate-and-sort on duplicate-heavy finite weights (seen at `top_n=4`). Fix verified: seed the spur search with the root's accumulated key → 0/16680. |
| D3 | defect | `braket_to_cirq` orders merged readout by instruction order, ignoring `Measure`'s classical bit index — a `from_ir` circuit with `b[2] = measure q[0]` comes out transposed. Same bug class as the moment-order bug this PR fixes in `cirq_to_pyquil`. |
| D4 | defect | `pytket_to_braket`'s mid-circuit guard misses a qubit measured into two different bits; one classical bit is silently dropped. |
| N5 | nit | `top_n=0` now returns 1 path (yield happens before the `yielded < k` check); main raised `ConversionPathNotFoundError`. |
| N6 | nit | A negative `edge_bias` (nothing validates it) silently breaks the search: 228/376 fuzz mismatches at bias −0.4, where the old enumerate-and-sort was exact. Worth `bias >= 0` validation in `Conversion._get_adjusted_weight`, since negative costs void the Dijkstra precondition. No anchorable line — the validation would go in `edge.py` (untouched). |
| N7 | nit | Multi-register QASM through the direct `cirq -> pyquil` edge orders registers alphabetically, while the `qasm2 -> pyquil` route uses DECLARE order — the two routes disagree on which qubit lands in bit 0. Probably unfixable from cirq alone (declaration order is lost); worth a comment. |

**Important context for D1/D2**: the shipped native graph has no weight-0 edges and no near-tie float collisions, so `transpile()` behavior on the default graph is exactly as documented — every claim in the PR body about the native graph checked out. D1/D2 matter for the same reason this PR replaced enumeration at all: `ConversionGraph` accepts arbitrary user conversions, and the PR's own tests (`test_ranking_holds_with_infinite_cost_edges`) currently give false confidence because their graphs are too small to trigger the failure.

**Suggested fixes** (both verified against the fuzz):
1. **D2 (and the decomposition half of D1)**: give `_cheapest_path` an optional `initial_key` and have `_k_cheapest_paths` pass the root's `(cost, hops, aliases)`. Every cost then accumulates left-to-right from the source — bit-identical to `_rank_key`'s `sum()` — and the spur search minimizes the *total* key, restoring Yen's optimality proof. Fuzz: 16680/16680 exact.
2. **D1's Dijkstra half**: map weight 0 to a large *finite* cost in `edge.py` (e.g. `1e6`) instead of `float("inf")`. Addition becomes translation-invariant again; ordering semantics shift from "all-inf paths tie, hops decide" to "fewer zero-weight edges wins, then the rest of the key" — arguably better, and it makes both the definition and the search agree. The two inf tests would need updating to the new semantics.

D3/D4 have small, local fixes (sort by `_target_index` with an instruction-order fallback; raise on re-measured qubit). Happy to push any of these.

qbraid/transpiler/graph.py
ryanhill1: Fixed in fe08ffcd (D1): weight 0 now maps to a large finite cost (1e6) in `edge.py`, restoring translation-invariance. The counterexample is pinned as `test_zero_weight_ties_rank_by_fewest_zero_weight_edges`, the two zero-weight tests are updated to the new semantics, and the generated-graph fuzz now includes 0.0 in its weight pool. 36,192 external fuzz comparisons match enumeration exactly.

qbraid/transpiler/graph.py
ryanhill1: Fixed in fe08ffcd (D2): `_cheapest_path` takes an `initial_key` and Yen's seeds each spur search with its deviation root's ranking key, so every cost is the same left-to-right sum `_rank_key` computes and the spur search minimizes the full candidate's key. The failing fuzz case is distilled into the deterministic `test_ranking_matches_enumeration_near_float_cost_ties`, and the generated-graph fuzz now also checks `top_n=8`.

qbraid/transpiler/graph.py
ryanhill1: Fixed in fe08ffcd (N5): `_k_cheapest_paths` returns immediately for `k <= 0`, so `find_top_shortest_conversion_paths(top_n=0)` raises `ConversionPathNotFoundError` again, matching main. Covered by `test_top_n_zero_finds_no_path`. Negative bias (N6) is also rejected with a `ValueError` in `Conversion._get_adjusted_weight`, covered by `test_negative_bias_raises`.

qbraid/transpiler/conversions/braket/braket_to_cirq.py
ryanhill1: Fixed in 494c669b (D3): measures are collected as `(index, qubit)` pairs and sorted by `getattr(operator, "_target_index", 0) or 0`; the sort's stability keeps instruction order for directly built Measures (which all default to 0). A rotation-permutation test (`test_braket_to_cirq_orders_readout_by_classical_index`) pins it, so an endianness flip can't masquerade as correct. Fails on the previous code.

qbraid/transpiler/conversions/pytket/pytket_extras.py
ryanhill1: Fixed in 494c669b (D4): the guard now checks the measured-qubit intersection before classifying the command, so a second Measure on a measured qubit raises `ProgramConversionError` like any other mid-circuit shape. Covered by `test_pytket_to_braket_remeasured_qubit_raises`.

qbraid/transpiler/conversions/cirq/cirq_to_pyquil.py
ryanhill1: Addressed in 494c669b (N7): documented in `_classical_bit_order`'s docstring — register-name order here vs DECLARE order on the qasm2 route, and why declaration order is unrecoverable from cirq keys.

ryanhill1:
Disposition of the two argus findings outside the diff (no inline thread to reply to):

- **`pytket_extras.py` — reused classical bits** (warning): **confirmed and fixed in 4c3d2b33.** Two qubits measured into one classical bit converted silently with two Braket result bits where the pytket register held one (later measurement overwrites in pytket semantics, unrepresentable in Braket). Now raises `ProgramConversionError`, with a regression test.
- **`tests/transpiler/pyquil/test_conversions_pyquil.py` — application import inside the `try/except ImportError`** (warning): **dismissed as pre-existing pattern.** The module's existing structure already puts `qbraid` imports (`circuits_allclose`, `cirq_to_pyquil`, `pyquil_to_cirq`) inside that block; the one line this PR added follows it. Restructuring the module's import guard is worth doing but belongs in its own change, applied consistently across the coverage test modules that share the pattern.

Of the 8 findings: 4 confirmed and fixed (subnormal-weight overflow, depth-cap bounding, reused classical bits, mapping assertion), 4 dismissed with evidence in the thread replies (cirq Qids are cross-type ordered; `Conversion.weight` is already the cost; cudaq output's comment header would fail a `startswith` check; import pattern pre-existing).

qbraid/transpiler/edge.py
```diff
@@ -166,8 +176,15 @@ def _get_adjusted_weight(self, weight: Optional[float] = None) -> float:
         if not 0 <= effective_weight <= 1:
             raise ValueError("Weight must be a float between 0 and 1, inclusive.")
 
+        if self._bias < 0:
+            # A negative bias makes edge costs negative, which silently breaks the
+            # shortest-path search (Dijkstra requires non-negative costs).
```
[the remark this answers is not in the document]
ryanhill1: Confirmed and fixed in 4c3d2b33. `Conversion(weight=5e-324)` produced cost `inf` — silently, since `np.log(1/5e-324)` overflows without a warning through numpy — which both reintroduced the inf pathology fixed in fe08ffcd and ranked a valid positive weight *worse* than weight 0 (1e6). Now `-np.log(w)`, which caps at ~745. Regression test asserts the subnormal cost is finite and below the zero-weight cost.

qbraid/transpiler/conversions/cirq/cirq_to_pyquil.py
```diff
@@ -39,6 +41,100 @@
...
+    QASM-derived circuits key single-qubit measurements ``c_0``, ``c_1``, ..., and that
+    suffix -- not the moment the operation happens to sit in -- is the bit position the
+    result belongs at: ``measure q[2] -> c[0]`` must land q_2 in bit 0 even though a
+    measurement on q_0 may appear earlier. Keys without an index fall back to qubit order,
+    matching the readout convention the Braket converters document.
+
+    A circuit measuring into multiple registers merges them in register-name order, since
+    cirq keys do not record declaration order. ``qasm2 -> pyquil`` concatenates registers
+    in DECLARE order instead (#1314), so the two routes can disagree on multi-register
+    circuits; ``transpile()`` normally routes around this edge.
```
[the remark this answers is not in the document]
ryanhill1: Dismissed: cirq `Qid`s are totally ordered across types by design — `min([cirq.NamedQubit("a"), cirq.LineQubit(0), cirq.GridQubit(0, 1)])` and `sorted(...)` over mixed types both work fine (verified on cirq 1.6.1; `Qid._cmp_tuple` includes the type name precisely so heterogeneous circuits sort). The fallback ordering cannot raise here.

tests/transpiler/test_conversion_graph.py
```diff
@@ -84,6 +89,13 @@ def bound_method_str(source, target):
     return f"<bound method Conversion.convert of ('{source}', '{target}')>"
 
 
+def _path_cost(graph: ConversionGraph, path: list) -> float:
+    """Total conversion cost of a path given as a list of bound Conversion.convert methods."""
+    aliases = _get_path_from_bound_methods(path).split(" -> ")
+    costs = {(conv.source, conv.target): conv.weight for conv in graph.conversions()}
+    return sum(costs[pair] for pair in pairwise(aliases))
```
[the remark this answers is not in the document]
ryanhill1: Dismissed — this misreads the cost model: `Conversion.weight` *is* the adjusted cost (`-log(w) + bias`, computed in `_get_adjusted_weight`), the same value `ConversionGraph` stores on each edge and sums in `graph._path_cost`. The helper therefore already computes exactly the production ranking quantity. The suggested change would double-apply the log transform, and `conv.bias` does not exist as a public attribute — it would raise `AttributeError`. (The property being named `weight` while returning a cost is genuinely confusing; renaming it is worth a follow-up outside this PR.)

qbraid/transpiler/conversions/cudaq/cudaq_to_qasm2.py
```diff
@@ -33,4 +34,9 @@
 @weight(1)
 def cudaq_to_qasm2(kernel: PyKernel) -> Qasm2StringType:
     """Converts a CUDA-Q kernel to QASM2."""
-    return cudaq.translate(kernel, format="openqasm2")
+    qasm = cudaq.translate(kernel, format="openqasm2")
+    # cudaq.translate reports failure by returning "{translation failed}"
+    # (with the error printed to stderr) instead of raising.
+    if "OPENQASM" not in qasm:
+        raise ProgramConversionError(f"cudaq.translate failed to produce QASM2: {qasm!r}")
+    return qasm
```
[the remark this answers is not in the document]
ryanhill1: Dismissed: real `cudaq.translate` output leads with a `// Code generated by NVIDIA's nvq++ compiler` comment before the `OPENQASM 2.0;` line, so `lstrip().startswith("OPENQASM 2.0;")` would reject *valid* translations. cudaq signals failure with the exact sentinel `"{translation failed}"`; the loose substring check distinguishes that sentinel from QASM, which is all this guard is for — anything malformed beyond that fails loudly in the downstream QASM parser, which is the fallback path working as designed.

qbraid/transpiler/graph.py
```diff
@@ -303,10 +304,154 @@ def remove_conversion(self, source: str, target: str) -> None:
...
+                    root[-1],
+                    target_id,
+                    frozenset(root[:-1]),
+                    banned_edges,
+                    initial_key=self._rank_key(root, alias_of),
+                )
+                if spur is None:
+                    continue
+                candidate = root[:-1] + spur
+                if candidate not in candidates and candidate not in accepted:
```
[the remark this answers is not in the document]
ryanhill1: Confirmed and fixed in 4c3d2b33, with a different mechanism than suggested: bounded DFS enumeration is O(V^max_depth) even when qualifying paths exist, so instead the depth cap now travels into every Dijkstra call as a node-count budget. Capped searches settle per (node, hops) state — under a budget, a costlier-but-shorter label is not dominated by a cheaper-but-longer one, which is why a naive hop-pruned Dijkstra would be inexact. Old behavior on a dense 12-node graph with max_depth=1 and no direct edge: still enumerating after 20s. New: <1ms. Verified equivalent to enumerate-filter-truncate on 23,775 fuzz comparisons over max_depth 1–4, and the `qualifies()` post-filter is gone entirely.

tests/transpiler/test_measurement_coverage.py
```diff
@@ -0,0 +1,166 @@
...
+    assert all(measures), "unparsable MEASURE instruction"
+    targets = {m.group(1) for m in measures}
+    assert targets == {register}, f"MEASURE targets {targets} != declared {register!r}"
+    indices = sorted(int(m.group(2)) for m in measures)
+    assert indices == list(range(NUM_QUBITS)), f"readout bits not distinct/complete: {indices}"
+
+
+@pytest.fixture(scope="module")
+def graph() -> ConversionGraph:
+    """Conversion graph shared across the sweep."""
```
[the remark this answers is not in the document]
ryanhill1: Applied in 4c3d2b33: the sweep now asserts the qubit-to-readout-bit mapping is the identity (the GHZ source measures q[i] into c[i]), not merely distinct-and-complete. All reachable pairs still pass, so no route currently permutes — and one that starts to will now fail this test rather than only the execution-based checks in test_measurement_semantics.py.

ryanhill1:
@coderabbit review

CHANGELOG.md
```diff
@@ -38,6 +40,8 @@ Writing an entry:
...
 
 ### Fixed
+- Measurement instructions now survive every conversion path. `cirq_to_braket` converts `MeasurementGate`s to Braket `Measure` instructions instead of silently dropping them (mid-circuit measurement raises `ProgramConversionError`, since Braket cannot represent it); `pytket_to_braket` re-appends the measurements that pytket's `tk_to_braket` discards; `braket_to_cirq` collects Braket's keyless terminal measurements into one keyed `MeasurementGate` instead of per-qubit `M('')` ops that downstream exports collapse; and `cirq_to_pyquil` merges terminal measurements into a single readout register instead of declaring one `BIT[1]` register per measurement key -- the fragmented form QCS rejects. The measured-circuit benchmark's known-failure list is now empty ([#1313](https://github.com/qBraid/qBraid/pull/1313))
+- `cudaq_to_qasm2` now raises `ProgramConversionError` when `cudaq.translate` fails -- cudaq reports failure by returning the literal string `"{translation failed}"` with the error printed to stderr, which the transpiler previously passed along as a successful conversion instead of falling back to another path ([#1313](https://github.com/qBraid/qBraid/pull/1313))
```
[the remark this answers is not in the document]
ryanhill1: Already addressed — the entry no longer makes the unqualified claim. Current text scopes it to exactly the covered routes:

> Measurement instructions now survive conversion between the program types the new measured-circuit benchmark covers (braket, cirq, pyquil, pytket, qiskit, and the qasm2/qasm3/openqasm3 dialects).

CUDA-Q and Qrisp are outside that list, as you noted, and the mid-circuit limitation is retained. No further change needed.

tests/transpiler/qrisp/test_coverage_from_qrisp.py
```diff
@@ -150,7 +150,11 @@ def is_package_installed(package_name: str) -> bool:
     return importlib.util.find_spec(package_name) is not None
 
 
-ALL_TARGETS = [("cirq", 0.95), ("pytket", 0.88), ("qiskit", 0.95)]
+# pytket is 0.85, not 0.88: MCRXGate's conversion is order-dependent (it passes only when
+# another qrisp coverage param has already run in the same process, warming qrisp's gate
+# synthesis). Cold, this param fails 4/25 = 0.84; under xdist the params land on arbitrary
+# workers, so the floor must accommodate the cold case or CI fails by scheduling lottery.
+ALL_TARGETS = [("cirq", 0.95), ("pytket", 0.85), ("qiskit", 0.95)]
```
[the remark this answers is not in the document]
ryanhill1: Already addressed, and by the route you suggested rather than a lowered floor — the revision this comment was written against is gone.

The test no longer decides pass/fail on the accuracy ratio at all. It asserts on failing gate **names**:

```python
unexpected = sorted(set(failures) - known_bad)
assert not unexpected, ...
```

So any gate failing outside `KNOWN_BAD_CONVERSIONS` fails the test regardless of what the rate is, and the message names which conversion regressed. `baseline` survives only as the parametrize id and as documentation of expected coverage. There is no 0.85 floor and no 0.01 allowance to absorb a real break.

The four broken qrisp -> pytket conversions are named explicitly (`CPGate`, `MCRXGate`, `RGate`, `U1Gate`) rather than averaged away, and the list is kept from rotting by the converse assertion: a known-bad gate that starts passing must be removed, or the test fails.

`MCRXGate` is the one carve-out from that converse check, because it fails only on some paths — the direct qrisp -> pytket edge produces a non-equivalent circuit while a route through an intermediate does not, and which one the graph picks depends on what else is imported. That is the cold-process nondeterminism you flagged; it is contained rather than hidden, and the underlying conversion is still wrong and tracked separately.

PR #1314, opened 2026-08-02
fix(rigetti): concatenate readout registers in DECLARE order, not sorted order

### LINKED ISSUE #1308: RigettiJob._parse_results orders readout registers lexicographically, silently corrupting multi-register bitstrings
## Summary

`RigettiJob._parse_results` concatenates readout registers in `sorted()` (lexicographic) order. Quil declaration order is not preserved, so for any program declaring more than one classical register whose names do not happen to sort into declaration order, **every returned bitstring has its bits in the wrong positions**. No error is raised; the job completes and the user receives plausible but incorrect measurement counts.

This has already happened in production.

## The code

`qbraid/runtime/rigetti/job.py`, in `_parse_results`:

```python
register_arrays = []
for reg_name in sorted(declared_registers):          # <-- lexicographic
    matrix = register_map.get_register_matrix(reg_name)
    if matrix is None:
        continue
    register_arrays.append(matrix.to_ndarray().astype(int))

all_bits = np.hstack(register_arrays)
```

`declared_registers` is a `set` built from `self._ro_sources` keys, so ordering information is already gone by this point. `sorted()` is the only thing imposing an order, and it is not the program's.

## Confirmed production occurrence

Job `rigetti:rigetti:qpu:cepheus-1-108q-8ce1-qjob-69f50b0600d5b375d6a597fe` (2026-05-01, status `COMPLETED`). Submitted program, format `quil`:

```
DECLARE ro BIT[2]
DECLARE r1 BIT[2]
H 0
CNOT 0 1
CNOT 0 2
CNOT 0 3
MEASURE 0 ro[0]
MEASURE 1 ro[1]
MEASURE 2 r1[0]
MEASURE 3 r1[1]
```

Declaration order is `ro`, `r1`. `sorted(["ro", "r1"])` returns `["r1", "ro"]`, so the bitstring was assembled as:

```
r1[0] r1[1] ro[0] ro[1]      (what was returned)
ro[0] ro[1] r1[0] r1[1]      (what the program declares)
```

Qubits 2 and 3 were reported in positions 0 and 1, and qubits 0 and 1 in positions 2 and 3. A second job with the same shape ran on 2026-05-04.

Worth noting how easily this hides: the circuit above prepares a GHZ state, whose only outcomes are `0000` and `1111`. Both are invariant under the swap, so the corruption is undetectable in the results themselves. A user with an asymmetric circuit would get silently wrong counts.

## Scope

Of 875 Rigetti jobs carrying `ro_sources`, 9 declare more than one register:

| date | registers (as stored) | sorted | affected |
| --- | --- | --- | --- |
| 2026-05-01 | `ro`, `r1` | `r1`, `ro` | **yes** |
| 2026-05-04 | `ro`, `r1` | `r1`, `ro` | **yes** |
| 2026-05-27 | `m0`..`m4` | `m0`..`m4` | no |
| 2026-07-17 | `m0`..`m5` | `m0`..`m5` | no |
| 2026-07-18 | `m0`..`m9` | `m0`..`m9` | no |
| 2026-07-29 (x4) | `m0`..`m7` | `m0`..`m7` | no |

The `m<N>` jobs are correct by luck: for a single digit, lexicographic order equals numeric order. A program with 11 or more such registers would break, because `sorted()` places `m10` between `m1` and `m2`. No production job has hit that yet.

Single-register programs (the overwhelming majority) are unaffected.

## Why this is not a one-line fix

The correct order is Quil declaration order, and **it is not recoverable from `_ro_sources`**. That dict comes back from `translate()` in Rust `HashMap` iteration order, which is neither declaration order nor index order. Real production data, from the job above:

```json
{"q0_unclassified": "q0_unclassified", "q1_unclassified": "q1_unclassified",
 "c[1]": "q1_classified", "c[0]": "q0_classified"}
```

Note `c[1]` precedes `c[0]`. Replacing `sorted()` with dict insertion order would trade one wrong answer for another.

Fixing this properly requires threading the compiled program's `DECLARE` order through from submit time. Open PR #1302 (expose the quilc-compiled program on `RigettiJob`) plausibly provides the vehicle, since the compiled Quil contains the `DECLARE` statements in order.

It also needs a decision on the contract: should multiple registers be concatenated into one bitstring at all, or should `GateModelResultData` carry per-register structure? Concatenation is inherently lossy about which bits belong to which register.

## Suggested handling

1. Decide the multi-register contract before implementing.
2. Until then, consider **failing loudly** when `len(declared_registers) > 1`, rather than silently emitting an arbitrary order. A clear error beats wrong numbers, and it affects a very small number of jobs.
3. The two 2026-05 jobs returned incorrect data to a user. Worth deciding whether that needs communicating.

## Related

- Found while adversarially reviewing #1301 (qasm2 to pyQuil conversion). Before that PR, the Cirq route fragmented readout into `m0..mN` registers, which made this reachable far more often; #1301 emits a single `ro` for OpenQASM input and so reduces exposure, but the raw-Quil path is unaffected.
- The `if matrix is None: continue` immediately above is separately unreachable dead code (`declared_registers` and the register map are both derived from `_ro_sources`, and `qcs_sdk` raises rather than omitting a register). Not a bug, but a cleanup candidate.
- Test fixtures in `tests/runtime/rigetti/test_rigetti_job.py` do not include the `q<N>_unclassified` entries that `translate()` actually returns, so they do not mirror production `ro_sources`.

ryanhill1:
Fixes #1308.

## What was wrong

`RigettiJob._parse_results` concatenated readout registers in `sorted()` order. Quil declaration order is not lexicographic order, so a program declaring `ro` then `r1` returned every bitstring with its halves swapped — silently, since the job completes and the counts look plausible. Two production jobs (2026-05-01, 2026-05-04) returned corrupted data this way; the `m0..m9` jobs in production were correct only because single-digit suffixes happen to sort numerically.

## The contract decision

Multi-register bitstrings concatenate in the **program's `DECLARE` order** — the order a Quil programmer reads in their own source. That order is recovered from the compiled program stored on the job at submission (#1302), which is exactly the program that was translated and executed. `ro_sources` cannot supply it: it arrives in Rust HashMap iteration order (production data shows `c[1]` before `c[0]`), so dict order would just be a different wrong answer.

## Fail loudly where the order is unrecoverable

- Multi-register jobs **rehydrated by ID** have no compiled program; `_parse_results` now raises `RigettiJobError` explaining why, instead of emitting an arbitrary order. Per #1308's production scan this affects a small minority of jobs (9 of 875 declared multiple registers).
- A register present in `ro_sources` but missing a `DECLARE` statement, or absent from the register map, now raises instead of being silently skipped (removes the unreachable `if matrix is None: continue` branch flagged in the issue).
- Single-register jobs — the overwhelming majority — need no order and parse exactly as before, with or without the compiled program.

## Tests

The sorted-order test that pinned the bug is replaced by a declaration-order test reproducing the production shape (`ro`/`r1`, asymmetric data so a wrong order changes the counts), including the `q<N>_unclassified` entries `translate()` really returns in `ro_sources` (per the issue's fixture-realism note). New tests cover the rehydrated-job refusal, the missing-DECLARE error, and the single-register no-program path. Full `tests/runtime/rigetti`: 258 passed.

Not addressed here: whether the two 2026-05 jobs' users should be notified (needs a product decision), and per-register (non-concatenated) result structure, which would be an API addition rather than a bug fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Rigetti measurement results so multiple readout registers are combined in their declared program order.
  * Added clearer errors when compiled program metadata or expected register data is unavailable.
  * Preserved existing behavior for single-register results.
  * Ignored unrecognized readout entries during result parsing.

* **Documentation**
  * Added an unreleased changelog entry describing the measurement result fix.


ryanhill1:
Live QPU verification (per CONTRIBUTING's "verify against the real service at least once"):

Submitted the two-register discriminator to Cepheus-1-108Q through `RigettiDevice.run()` on this branch — `DECLARE ro BIT[2]` then `DECLARE r1 BIT[2]`, with `X` on both qubits measured into `ro`, so correct ordering must read `1100` and the pre-fix `sorted()` ordering would deterministically read `0011`.

Job `51709b23-9fd1-4cc6-a095-892893df1c34` (2026-08-03):

```
compiled DECLAREs (order of record): ['DECLARE ro BIT[2]', 'DECLARE r1 BIT[2]']
ro_sources (HashMap order): ['q3_unclassified', 'q1_unclassified', 'q0_unclassified',
                             'ro[1]', 'q2_unclassified', 'ro[0]', 'r1[0]', 'r1[1]']
counts: {'1100': 1}
```

The live `ro_sources` arrived scrambled exactly as the issue describes (`ro[1]` before `ro[0]`, `q<N>_unclassified` entries interleaved — matching the test fixture), and `_register_order` recovered the DECLARE order and emitted the `ro`-half first.

Caveat: 1 of 100 shots returned — the known artifact of the `ConnectionStrategy.EndpointAddress` path, which is the only route reachable from this network (the default Gateway strategy times out locally; production traffic uses Gateway and returns full counts). Shot count is orthogonal to register ordering: the concatenation applies identically per shot, and the multi-shot multi-register path is covered by unit tests built on this same payload shape.

PR #1321, opened 2026-08-08
fix(runtime): pad IonQ counts to the submitted register width

ryanhill1:
### Problem

IonQ returns probabilities keyed by decimal state index. `IonQJob` handed those ints to `normalize_data`, which renders a key as `bin(key)` and then pads only to the widest key present in the dict. Two consequences:

- `x q[0]` on a six-qubit register returns `{"1": N}`. Every other qBraid route — Braket, OpenQuantum, and qir-sv after qBraid/docker-stacks#420 — returns `{"000001": N}`.
- Key width tracks whichever outcomes a run happened to sample, not the register. The same circuit gives `{"1": 100}` on a clean run and `{"000001": 90, "100001": 10}` on a noisy one. Callers that slice by bit position get a different answer per run.

Reported by a customer after ~35 hardware jobs; the format difference across routes cost them debugging time.

### Width comes from `stats.qubits`

IonQ echoes the submitted register size back in the job stats. I checked that it is the *declared* width, not the used-qubit count — a 6-qubit register with a single `x q[0]` reports `stats.qubits: 6`. That is the only server-side source, and it survives jobs rehydrated by ID, which the submitted payload does not.

When the field is absent, `_rekey_to_bitstrings` falls back to the widest observed outcome — i.e. today's behaviour. I'd rather degrade than raise on a response shape I can't fully enumerate, but it does mean the bug is still reachable if IonQ ever stops sending stats.

### Bit order is already right

Qubit 0 is the least significant bit of IonQ's index, so `format(v, f"0{w}b")` puts it rightmost, matching the convention `CLIENT_QA_RESPONSE.md` documents. Only padding changes here; no reversal.

### Verified

End to end against the live IonQ simulator — `x q[0]`, `x q[5]`, `x q[0];x q[2]` on six qubits now return `000001`, `100000`, `000101`, with probabilities keyed to match.

Unit tests: 7 added, 5 of which fail on the current code. The two that pass either way are the deliberate fallback and `x q[5]`, where the unpadded key coincides with the padded one — that coincidence is why this went unnoticed.

`tests/runtime` is otherwise unchanged: 9 failures, all pre-existing on `main` (6 `compiled_program`, 3 Azure remote tests needing credentials).


ryanhill1:
Superseded by #1322, which folds this padding fix in as its first commit and then reverses the key order as part of the platform-wide move to qubit-0-first. Closing to avoid merging a fix that #1322 immediately rewrites.

PR #1330, opened 2026-08-15
Preserve circuit widths in batch counts

### LINKED ISSUE #1323: BatchResult.data.get_counts() pads all circuits to the batch's widest register
### Summary

For a batch job whose circuits have **different register widths**, `BatchResult.data.get_counts()` zero-pads every circuit's bitstrings to the widest register in the batch. A 1-qubit circuit is reported as having a 3-bit register, and the aggregate view disagrees with the per-circuit view of the same data.

### Reproduction (no network required)

```python
from qbraid.runtime.postprocess import normalize_data

# a 2-qubit Bell, a 3-qubit GHZ, and a 1-qubit X
batch = [{'00':18,'01':24,'10':24,'11':34}, {'000':14,'111':15}, {'0':3,'1':97}]

for c in normalize_data(batch):
    print(c)
```

```
{'000': 18, '001': 24, '010': 24, '011': 34}   # was 2-bit
{'000': 14, '111': 15}                         # unchanged
{'000': 3, '001': 97}                          # was 1-bit
```

Expected: each circuit keeps its own register width.

### Where

`normalize_data()` (`qbraid/runtime/postprocess.py:194`) routes list input through `normalize_batch_bit_lengths()` (same file, line 26), which computes one `max_bit_length` across **all** dicts in the batch and `zfill`s every key to it.

The function is correct for its original purpose — a single circuit whose backend drops leading zeros, so `'0'` and `'11'` need reconciling to a common width. Applied across independent circuits, the shared width is not a property they have.

### Impact

The aggregate and per-circuit accessors disagree for the same job:

```python
result.data.get_counts()[2]        # {'000': 3, '001': 97}
result.results[2].data.get_counts()  # {'0': 3, '1': 97}
```

`results[i]` is correct — it takes the single-dict `format_data` path and never hits batch normalization. Both are documented access patterns in the [batch jobs guide](https://docs.qbraid.com/v2/sdk/user-guide/providers/native/jobs#batch-jobs), so the same job yields two different answers depending on which one a user reaches for. Anyone deriving qubit counts, marginals, or plot labels from the aggregate keys gets a register width the circuit never had.

Homogeneous batches are unaffected (`all_keys_same_length` short-circuits), which is why this stays hidden until circuits of differing widths are batched together.

### Observed on

`qbraid 0.12.3.dev20260716220927`, against a real batch on `qbraid:equal1:sim:bell-1` (`job.result()` → `BatchResult`, `num_circuits=3`). The server-side payload is correct — each `results[i].resultData.measurementCounts` carries its own width — so this is purely client-side post-processing.

### Suggested fix

Normalize within each circuit rather than across the batch — apply `normalize_bit_lengths` per dict in the batch path, leaving `normalize_batch_bit_lengths` for callers that genuinely want one uniform width.

danielgaskins:
## Summary of changes

`BatchResult` aggregates independent measurement dictionaries from its per-circuit results. Each dictionary is already passed through `format_data()`, which normalizes bit lengths within that circuit. `normalize_data()` then sent the full list through `normalize_batch_bit_lengths()`, padding every circuit to the widest register.

Removing the batch-wide call is sufficient because the per-circuit formatting has already happened. Single-circuit normalization remains unchanged, while heterogeneous batches preserve the register width for each circuit.

The regression test exercises an actual `BatchResult` and verifies that its aggregate and per-circuit views agree. Additional assertions cover mixed-width batches with zero-filled and decimal counts.

Closes #1323

## Testing

- `tox -e unit-tests`
- `tox -e format-check`
- `tox -e docs`

## Summary by CodeRabbit

* **Bug Fixes**
  * Batch results now preserve each circuit’s individual measurement-register width.
  * Removed unnecessary padding of result counts to the widest circuit.
  * Reported measured-qubit widths and per-result counts are now accurate for mixed-width batches.

* **Documentation**
  * Added a changelog entry describing the corrected batch result formatting.


danielgaskins:
@coderabbitai review

tests/runtime/test_result.py
```diff
@@ -284,6 +284,24 @@ def test_batch_normalized_counts():
...
+    ]
+    results = [
+        Result("device_id", f"job_id_{index}", True, GateModelResultData(measurement_counts=item))
+        for index, item in enumerate(counts)
+    ]
+    batch_result = BatchResult("device_id", "batch_job_id", True, results)
+
+    assert batch_result.data.get_counts() == counts
+    assert [result.data.get_counts() for result in batch_result.results] == counts
+    assert batch_result.data.to_dict()["num_measured_qubits"] == [2, 3, 1]
```
ryanhill1, line 302: **N1:** this covers the default path only — `include_zero_values=True` and `decimal=True` with mixed widths both changed here, and neither is guarded. Please add coverage for both.

`decimal` is the important one: with the `if decimal` early return gone, anything that reintroduces a string-normalizing call in this function raises `TypeError: object of type 'int' has no len()` on every decimal batch, and no test catches it — `test_decimal_get_counts` passes a single dict, so it never reaches the batch path.

```python
assert normalize_data(counts, decimal=True) == [
    {0: 18, 1: 24, 2: 24, 3: 34}, {0: 14, 7: 15}, {0: 3, 1: 97}
]
```

CHANGELOG.md
```diff
@@ -40,6 +40,9 @@ Writing an entry:
 ### Removed
 
 ### Fixed
+- Fixed batch result counts padding every circuit to the widest register. Each circuit now
+  keeps its own measurement register width when using `BatchResult.data.get_counts()`
+  ([#1323](https://github.com/qBraid/qBraid/issues/1323)).
```
ryanhill1: **N2:** `CONTRIBUTING.md` says to link your PR in the changelog entry — please link #1330 here rather than the issue.

ryanhill1 (commented):
Thanks — reproduced #1323 locally and the fix is correct. Aggregate and per-circuit views now agree, and the new test genuinely fails on unpatched `main`.

One thing worth stating in the PR description: deleting the call rather than replacing it per-circuit (as the issue suggested) is safe because `format_data` already normalizes each dict at `postprocess.py:135`. Anyone comparing the PR to the issue will otherwise read this as under-delivering.

Two small fixes inline.

tests/runtime/test_result.py
danielgaskins: Added both cases to the mixed-width regression test. The zero-filled assertion keeps each circuit at its own width, and the decimal assertion covers the batch path with integer keys.

CHANGELOG.md
danielgaskins: Updated the changelog entry to link #1330.

ryanhill1 (approved):

ryanhill1:

PR #1331, opened 2026-08-18
feat(rigetti): add opt-in initial_rewiring runtime option

TheGupta2012:
## Summary of changes

Adds an opt-in `initial_rewiring` key to `RigettiDevice.run(runtime_options=...)`, selecting quilc's `INITIAL_REWIRING` strategy (`NAIVE`, `PARTIAL`, `GREEDY`, `RANDOM`).

```python
device.run(program, shots=100, runtime_options={"initial_rewiring": "GREEDY"})
```

Omitting the key leaves the program untouched, so quilc's default still applies and no existing caller changes.

### Why

quilc chooses the logical-to-physical qubit mapping with its `INITIAL_REWIRING` strategy. That strategy is a Quil **pragma**, not a quilc server flag, so it cannot be set on the compiler deployment — only by rewriting the program. The SDK had no way to set it, so callers always got quilc's `PARTIAL` default.

`PARTIAL` searches for a placement at a cost that is *independent of circuit depth* and grows sharply with qubit count. Benchmarked against the live Cepheus-1-108Q ISA (12x9 lattice, 107 live qubits, 193 CZ pairs, real per-edge fidelities), depth-1 compile under `PARTIAL` takes 2.1s at 36 qubits, 8.8s at 54, and ~128s at 107 — roughly O(n^3.8). At 107 qubits that is ~70% of the runtime-api's 180s `DEFAULT_COMPILER_TIMEOUT_S` spent before any useful work.

Measured alternatives at 107 qubits, depth 1:

| circuit | `PARTIAL` (today) | `GREEDY` | `NAIVE` |
|---|---|---|---|
| arbitrary qubit pairs | 127.8s → 53 CZs | **4.95s → 53 CZs** | 117.4s → 785 CZs |
| pairs already on lattice edges | 130.7s | 6.3s | **4.74s, no SWAP overhead** |

`GREEDY` is ~26x faster for byte-identical output, and `NAIVE` is ~27x faster on pre-mapped circuits *and* emits fewer gates.

The win narrows at depth ≥ 2, where `PARTIAL` costs about the same as `GREEDY` but produces better output (299 vs ~640 CZs). That is why this is opt-in rather than a changed default: picking one strategy for every circuit would trade output quality for speed on deeper programs.

### Behavior

- A program that already declares the pragma keeps its own value and logs a warning. Two `INITIAL_REWIRING` pragmas in one program are rejected by quilc.
- An unrecognized strategy raises `ValueError` naming the supported set, rather than surfacing as an opaque quilc rejection of the whole program.
- The pragma does not need to be the literal first line. pyquil's `out()` emits `DECLARE` ahead of it, and quilc honors it there (verified against quilc 1.26.0).

### Implementation note

The strategy is carried through *every* branch of `_resolve_compiler_options`, not only the first. It is not a `CompilerOpts` field, so `run()` publishes it with `options=None` whenever it is the only quilc key given — the common case — and resolving on `options` alone dropped it exactly there. Two tests guard that path.

### Testing

23 new tests covering parsing and validation, program rewriting, and the option reaching `compile_program` through `transform()`. Full `tests/runtime/rigetti/` suite passes (282 passed). Both regression guards were confirmed to fail against the pre-fix resolve path.

### Context

Companion to qBraid/qbraid-infrastructure#736, which moves the prod quilc VM off a shared-core machine type. That change is worth roughly 2x on compile latency; this one is worth ~26x. The two are independent.


ryanhill1:
Verified this end to end and it holds up well: the 282-test suite, lint, and mypy baseline are all green; both regression guards genuinely fail against the pre-fix resolve path (checked by mutation); and against the live Cepheus-1-108Q ISA with quilc 1.26.0, `transform()` with `initial_rewiring="GREEDY"` compiles with the pragma correctly placed after `DECLARE`, and the 54-qubit depth-1 timing reproduces your table almost exactly — 8.8s under default `PARTIAL` vs 1.5s under `GREEDY`. The parse-time `ValueError` also fires before anything reaches the QPU, and an in-program bad strategy produces exactly the opaque quilc rejection this design avoids.

One correction, pushed as c81b746: quilc 1.26.0 **accepts** a program with two `INITIAL_REWIRING` pragmas (both echo through compilation) rather than rejecting it. The dedupe guard is still right — for a stronger reason: quilc applies the pragma positionally, so prepending a device-level pragma ahead of the author's would silently change which strategy governs their program instead of failing loudly. The commit rewords the two docstrings that cited the rejection; no behavior or assertions changed. The same sentence appears in the PR description's Behavior section — worth the same tweak there, which I've left to you.

ryanhill1 (approved):

PR #1299, opened 2026-07-29
feat(runtime): add QPerfect (MIMIQ) provider

TheGupta2012:
## Summary

Adds a direct **QPerfect (MIMIQ)** provider giving qBraid users access to the MIMIQ cloud emulator, following the AQT/QUDORA integration pattern.

## What ships

- **`qbraid/runtime/qperfect/`** — `QPerfectProvider` / `QPerfectDevice` / `QPerfectJob`, driven by the `mimiqcircuits` + `mimiqlink` SDK. Token auth via `MimiqConnection`; a single `mimiq-emulator` device whose `status()` reflects connection health; MIMIQ submit options (`algorithm`, `nsamples`, `bonddim`, `entdim`, `seed`, `bitstrings`, `noisemodel`, ...) forwarded through `submit`. Batched (list) submission supported.
- **Transpiler node** — registers `mimiqcircuits.Circuit` as the `mimiq` node and adds a `qiskit -> mimiq` converter that delegates to the maintained **`mimiq-qiskit`** package. Any source reaches MIMIQ through the qiskit hub, while a native `mimiqcircuits.Circuit` passes through with noise / special ops preserved.
- **`MimiqProgram`** wrapper for `load_program` / `QuantumDevice.validate` qubit-count introspection.
- **Packaging** — `qperfect` extra (`mimiq-qiskit`), provider / job / program entry-points, the `qbraid/runtime` lazy registry, and the `docs/api/qbraid.runtime.rst` autodoc entry.

## Testing

- Mocked provider / device / job tests + a `qiskit_to_mimiq` converter test (mock `mimiq-qiskit`); `tests/runtime/test_loader.py` updated for the new provider.
- `black` / `isort` / `pylint` clean; `pytest tests/runtime/qperfect` green.

## Notes

- Install with `pip install "qbraid[qperfect]"` (pulls `mimiqcircuits` + `qiskit>=2.0`).
- Draft pending live end-to-end verification against MIMIQ with credentials.

PR #1332, opened 2026-08-18
fix(qperfect): use native mimiqcircuits alias and unblock batch submission

ryanhill1:
## Summary of changes

Addresses the two review comments on #1299, plus a batch submission bug found testing against the live MIMIQ cloud.

### Program type registered natively, under its derived alias

`mimiqcircuits` now goes through `dynamic_type_registry` with a `_get_class` branch, so it lands in `NATIVE_REGISTRY` like every other type with a wrapper in `programs/gate_model/`. The manual `dynamic_non_native["mimiq"]` override is gone and the alias derives from the module name.

That forces renames, which is the part worth a look: the transpiler derives conversion node names by splitting the function name on `_to_` (`conversions/__init__.py:82`), so the edge only becomes `qiskit -> mimiqcircuits` if the function is renamed too. Following the existing convention that a `gate_model` module is named for its alias:

- `gate_model/mimiq.py` → `mimiqcircuits.py`
- `qiskit_to_mimiq` → `qiskit_to_mimiqcircuits`
- `ProgramSpec(MimiqCircuit, alias="mimiq")` → `ProgramSpec(MimiqCircuit)`

`gate_model/qrisp.py` sets the precedent for a module shadowing a same-named third-party package.

### Batch submission could never succeed

`submit()` defaulted `algorithm` to `"auto"` for every request, but MIMIQ rejects `"auto"` in batch mode, so every list submission failed against the live cloud:

```
ValueError: The 'auto' algorithm is not supported in batch mode.
Please specify 'mps' or 'statevector' for batch executions.
```

The default now applies only to a single circuit; a batch without an explicit `algorithm` raises with an actionable message. Open to instead picking `"statevector"` as an implicit batch default — I chose to raise because the choice changes both accuracy and cost.

The existing `test_submit_batch` passed because it mocks the connection, so it never saw the constraint. Added a regression test.

### CONTRIBUTING compliance

Audited both PRs against `CONTRIBUTING.md`; three rules weren't met, all in code #1299 introduces.

- **Fail loudly on missing data.** `CONTRIBUTING` names this anti-pattern down to the variable name — index `_STATUS_MAP` directly rather than `.get(status, UNKNOWN)`. mimiqlink publishes its full status vocabulary in `RequestInfo.STATUS_COLORS` and `_STATUS_MAP` already covers all five, so a miss means the vendor added a member. Origin's provider already does this.
- **PEP 585/604 in new code.** `Optional[X]`/`Union[X, Y]` → `X | None`/`X | Y`, matching pasqal and quantinuum. Existing modules left alone, per the same rule's no-retrofit clause.
- **Verify against the real service.** Added `test_qperfect_remote.py` behind the `remote` marker, which every other provider ships and qperfect did not.

### Loader

Added QPerfect `load_job`/`load_provider` overloads and the `TYPE_CHECKING` imports.

## Verification

7 remote tests pass against `mimiq.qperfect.io`: device discovery, status, single-circuit Bell, endianness on asymmetric circuits (X on q0 of 3 → `001`, X on q2 → `100`), and both batch paths. Full suite green (2179 passed), `tox -e format-check` clean at pylint 10.00/10.

## Caveats

- **Batch results are padded to the widest register in the batch**: a 1-qubit circuit returns `{'01': 200}` in a batch but `{'1': 100}` alone. That's MIMIQ behavior passed through unchanged — left alone here, but arguably worth normalizing per circuit.
- The `qperfect` extra is **not co-installable with myQLM**: `mimiqcircuits` pins `bitarray<3`, `bitstring` (via `qat-core`) needs `>=3`. Not addressable in this PR.


ryanhill1:
**TODO**: Also need to update README QPROGRAM_REGISTRY with new program type

I think we forgot to do this for AQT as well

TheGupta2012:
@ryanhill1 thanks for the 

> **TODO**: Also need to update README QPROGRAM_REGISTRY with new program type
> 
> I think we forgot to do this for AQT as well

Done for both, AQT in PR #1339

PR #1333, opened 2026-08-18
fix(transpiler): support Cirq ResetChannel in pyQuil conversion

TheMickeyDodger:
## Summary of changes

Adds support for converting Cirq `ResetChannel` operations to pyQuil `RESET` instructions.

Previously, Cirq circuits containing `cirq.reset` failed conversion because `ResetChannel` was not handled by the Cirq-to-Quil output path. This change maps qubit resets directly to Quil `RESET` on the corresponding target qubit.

The change also adds regression coverage verifying that:

- Cirq reset operations convert successfully to pyQuil,
- the reset targets the correct qubit,
- operation ordering is preserved in circuits containing resets,
- qutrit resets continue to fail explicitly because Quil `RESET` is qubit-only.

Existing conversion behavior is otherwise unchanged.


ryanhill1 (approved):

ryanhill1:

PR #1343, opened 2026-08-20
Add OpenSSF Scorecard workflow to publish a public security score

ryanhill1:
Adds the OpenSSF Scorecard action, which grades a repository against a set of
supply-chain security checks and publishes the result publicly.

`qBraid/qBraid` has no published Scorecard today (`api.securityscorecards.dev` returns
nothing for it), so there is currently no third-party measure of the project's security
posture. After the first run there will be a score at
`https://api.securityscorecards.dev/projects/github.com/qBraid/qBraid`, and a badge is
available if we want one in the README.

### What it does

Runs weekly and on pushes to `main`, checks things like branch protection, signed
releases, dependency pinning, token permissions, fuzzing and CI tests, uploads SARIF to
code scanning, and publishes the aggregate score.

### Notes

- `publish_results: true` is what makes the score publicly visible. It reports only the
  check results, not source.
- This complements the CodeQL default setup already enabled here. CodeQL analyzes code for
  vulnerabilities; Scorecard grades the project's practices around that code.
- Expect the initial score to be middling. Several checks will fail on things we already
  know about, including the long-lived PyPI token in `publish.yml` and the absence of
  signed releases. That is the point of establishing a baseline.

## Summary by CodeRabbit

* **Chores**
  * Added automated security posture analysis for branch protection changes, weekly runs, main-branch updates, and manual requests.
  * Security findings are now published to code scanning, with short-term result retention for review.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -8,7 +8,17 @@ on:
     branches: [ main ]
   workflow_dispatch:
 
-permissions: read-all
+# Four triggers can overlap, and Scorecard reports current repository state, so a
+# superseded run has nothing to contribute. Cancelling keeps the published SARIF and the
+# public API entry in trigger order.
+concurrency:
+  group: scorecard
+  cancel-in-progress: true
+
+# Least privilege by default: `analysis` names everything it needs below, and a job added
+# later gets nothing until it does the same. Job-level permissions replace this block
+# rather than adding to it.
+permissions: {}
 
 jobs:
   analysis:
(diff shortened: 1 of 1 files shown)


ryanhill1:
Addressed in e31e4802, along with three findings raised on the identical copy of this workflow in qbraid-algorithms#94.

**Pin to SHAs** — done, all four, each with its release in a comment. I resolved every SHA from its upstream repository rather than copying from review text, since an unverified SHA is the supply-chain risk that pinning exists to prevent. Worth noting the point about `security-events: write` and `id-token: write` is what makes this more than hygiene here.

**This surfaced a real bug.** `ossf/scorecard-action@v2` **does not resolve** — that repo publishes `v2.3.3`…`v2.4.4` with no `v2` tag or branch, so the step could never have run. Pinning to `2d114668… # v2.4.4` fixes it.

Also ported from #94:

- `permissions: read-all` → `permissions: {}`. Job-level permissions replace the workflow block rather than extending it, so `analysis` is unaffected while a job added later starts with nothing instead of repo-wide read.
- A `concurrency` group, so overlapping runs from the four triggers cannot publish SARIF out of order.
- A comment on `actions: read`. It is not in OSSF's documented permission set; it comes from the starter template, where it is documented for **private** repositories. This repo is public, so it is likely removable — I kept it and said so rather than dropping it, since removing a permission a check quietly depends on would degrade the score silently.

The two workflows are byte-identical again after this change.

PR #1351, opened 2026-08-24
fix(runtime): reject unrecognized job statuses

### LINKED ISSUE #1335: Remove UNKNOWN status fallbacks from IBM and AWS job status maps
`CONTRIBUTING.md` is explicit about lookup tables keyed on vendor data:

> The same applies to lookup tables keyed on vendor data: index directly (`_STATUS_MAP[status]`) so an unmapped value raises, rather than `.get(status, SOME_DEFAULT)`, which hides it.

Two providers still use the fallback:

- `qbraid/runtime/ibm/job.py:99` — `IBM_JOB_STATUS_MAP.get(job_status, JobStatus.UNKNOWN)`
- `qbraid/runtime/aws/job.py:71` — `AWS_TASK_STATUS_MAP.get(state, JobStatus.UNKNOWN)`

Both maps are already complete against their vendor's declared vocabulary, so the fallback is unreachable for any status the SDK actually emits:

- **AWS** — botocore's Braket API model declares `QuantumTaskStatus = ['CREATED', 'QUEUED', 'RUNNING', 'COMPLETED', 'FAILED', 'CANCELLING', 'CANCELLED']`, exactly the 7 keys in `AWS_TASK_STATUS_MAP`. Clean removal.
- **IBM** — `qiskit_ibm_runtime` declares `JobStatus = Literal["INITIALIZING", "QUEUED", "RUNNING", "CANCELLED", "DONE", "ERROR"]`; `IBM_JOB_STATUS_MAP` covers all six plus a legacy `VALIDATING`.

One caveat before doing IBM: `RuntimeJobV2._status_from_job_response` returns the raw `api_status` string when it isn't in `API_TO_JOB_STATUS`, so `status()` can yield a value outside its own declared `Literal` if IBM adds an API status. Removing the fallback makes that raise instead of silently reading `UNKNOWN` — which is the intent, but it is a real behavior change on live jobs, not a no-op cleanup. AWS has no equivalent leak.

`origin/job.py:58` and `qperfect/job.py` (as of #1332) show the pattern to copy: check membership and raise with the expected values listed.

Split out of #1334, which covers `loader.py` hygiene.

danielgaskins:
## Summary of changes

Closes #1335.

Raises clear errors when AWS Braket or IBM Runtime reports a job status outside the SDK’s known values. This prevents newly added vendor states from being silently exposed as `UNKNOWN`.

Adds public `status()` tests for both providers. The shared IBM job fixture now returns `DONE`, matching `RuntimeJobV2`.

## Testing

- `python -m pytest -q tests/runtime/aws/test_braket_runtime.py tests/runtime/ibm/test_qiskit_runtime.py`
- `python -m black --check qbraid/runtime/aws/job.py qbraid/runtime/ibm/job.py tests/runtime/aws/test_braket_runtime.py tests/runtime/ibm/test_qiskit_runtime.py`
- `python -m ruff check qbraid/runtime/aws/job.py qbraid/runtime/ibm/job.py tests/runtime/aws/test_braket_runtime.py tests/runtime/ibm/test_qiskit_runtime.py`
- targeted mypy check

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ Writing an entry:
 ### Removed
 
 ### Fixed
+- Fixed AWS and IBM jobs silently returning `UNKNOWN` when a vendor reports an unrecognized status ([#1351](https://github.com/qBraid/qBraid/pull/1351)).
 - Fixed Open Quantum provider and job discovery through `get_providers()`, `load_provider()`, and `load_job()` ([#1347](https://github.com/qBraid/qBraid/pull/1347))
 - Fixed measurement results coming back permuted when a Cirq circuit reaches pyQuil through `qasm2`. Cirq declares one `creg` per measurement key in scheduling order, so a measurement on an idle qubit was declared first; flattening those registers into one readout region then followed declaration order and moved qubit 2's result into bit 0. Registers are now declared in key order ([#1345](https://github.com/qBraid/qBraid/pull/1345))
 - Fixed Cirq → pyQuil conversion failing with `ProgramConversionError` for circuits containing `cirq.reset` / `cirq.ResetChannel`. Resets now convert directly to Quil `RESET` on the target qubit; qudit resets (dimension > 2) still raise, since Quil `RESET` is qubit-only ([#1333](https://github.com/qBraid/qBraid/pull/1333))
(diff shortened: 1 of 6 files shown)


ryanhill1 (commented):
Thanks for the PR! A couple failing tests here. After those are fixed this is g2g

danielgaskins:
Sure thing, the AHS result test was using an unset `state()` mock, so the stricter status check saw the mock object as an unknown AWS state. I set it to `COMPLETED`, which matches the metadata already used by the test. The focused AWS AHS, AWS runtime, and IBM runtime suites now pass with 107 tests.

ryanhill1 (approved):

PR #1373, opened 2026-09-05
fix(transpiler): report unresolved Cirq parameters

### LINKED ISSUE #1363: transpile() of a parameterized cirq circuit raises raw numpy/sympy TypeErrors instead of a named unsupported-input error
Converting a cirq circuit that contains a free `sympy.Symbol` fails with low-level `TypeError`s raised from inside numpy and sympy, on every available path. The errors don't name the actual problem (a symbolic parameter), so the caller has to read a numpy ufunc message to work out what they did wrong.

I want to be precise about the scope: **the report is about the error surface, not about whether symbolic circuits should be supported.** Refusing them may well be correct, since Braket wants concrete angles.

### Reproduction

```python
import cirq, sympy
from qbraid.transpiler import transpile

theta = sympy.Symbol("theta")
circuit = cirq.Circuit([
    cirq.rx(theta).on(cirq.LineQubit(0)),
    cirq.H.on(cirq.LineQubit(0)),
    cirq.CNOT.on(cirq.LineQubit(0), cirq.LineQubit(1)),
])

transpile(circuit, "braket")
```

### Actual

```
ProgramConversionError: Failed to convert 'cirq' to 'braket' due to the following error(s):

Conversion cirq -> braket failed due to exception raised while converting from 'cirq'.
TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be
safely coerced to any supported types according to the casting rule ''safe''

Conversion cirq -> qasm2 -> qasm3 -> braket failed due to exception raised while converting from 'cirq'.
TypeError: unsupported format string passed to Mul.__format__

Conversion cirq -> qasm2 -> pytket -> braket failed due to exception raised while converting from 'cirq'.
TypeError: unsupported format string passed to Mul.__format__
```

`transpile(circuit, "qasm2")` and `transpile(circuit, "qasm3")` fail the same way.

### Expected

Either the parameter is handled, or the failure names it — something like "cannot convert a parameterized circuit to `braket`: bind or resolve `theta` first". `cirq.is_parameterized(circuit)` already answers this cheaply, so a check at the entry point would replace three confusing internal errors with one clear one.

### Environment

- qbraid `0.13.0.dev` (current `main`) — reproduced
- qbraid `0.12.2` (latest PyPI) — reproduced
- cirq `1.4.1`, Python 3.12, macOS

### Context

Found while building a differential-testing harness for SDK conversion layers, which also gives me some good news worth passing on: the same harness ran **192 round trips** through `braket`/`qasm2`/`qasm3` × `LineQubit`/`NamedQubit`/`GridQubit`/offset-`LineQubit` × 1–4 qubits × 4 seeds and found **zero** unitary mismatches, with the comparison metric calibrated against known-equal and known-unequal pairs first. Non-parameterized conversion looks solid; this is the one rough edge it surfaced.

*Disclosure: this report was prepared with AI assistance. Every command above was executed and the output is quoted verbatim.*

danielgaskins:
## Summary

Parameterized Cirq circuits currently reach numeric conversion code and leak low-level NumPy or SymPy `TypeError`s. This change checks for free parameters at the Braket and OpenQASM 2 boundaries, then raises `ProgramConversionError` with the sorted parameter names and guidance to resolve them. Cirq to PyQuil remains unchanged.

The tests cover both direct conversions and `transpile()` paths to Braket, QASM 2, and QASM 3.

Closes #1363

## Testing

- `python -m pytest -q tests/transpiler/cirq tests/transpiler/test_transpiler.py -p no:cacheprovider`
- 371 passed, 20 skipped
- Import order, Black, Ruff, license headers, and focused Pylint checks


tests/transpiler/cirq/test_conversions_cirq.py
```diff
@@ -161,3 +163,30 @@ def test_cirq_to_qasm2_leaves_unindexed_creg_names_in_place():
...
+        cirq.rx(theta).on(cirq.LineQubit(0)), cirq.ry(alpha).on(cirq.LineQubit(1))
+    )
+
+    with pytest.raises(
+        ProgramConversionError,
+        match=(
+            r"Cannot convert a Cirq circuit to OpenQASM 2 with unresolved parameters: "
+            r"alpha, theta\. Resolve the parameters before conversion\."
+        ),
+    ):
```
ryanhill1, line 181: **B1 (blocking).** This parametrized test passes for `braket`/`qasm3` but **fails for `qasm2`** when `pytket-cirq` is installed (it is, via `requirements-dev`): `transpile` reaches the target through the unguarded `cirq→pytket→qasm2` route, which silently returns invalid QASM (`rx((theta/pi)*pi) q[0];`) so `pytest.raises` never fires. CI will fail here.

Also note the `match=` here is a substring, so `[braket]`/`[qasm3]` pass even though their aggregated error still contains a raw `TypeError` from the pytket hop (D1) — the test can't tell a clean rejection from a mixed one. Consider asserting the raw error is *absent*, not just that the named one is present.

qbraid/transpiler/conversions/cirq/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Args:
+        circuit: Cirq circuit to validate.
+        target: Name of the concrete target format.
+
+    Raises:
+        ProgramConversionError: If the circuit contains unresolved parameters.
+    """
+    parameters = sorted(cirq.parameter_names(circuit))
+    if parameters:
```
ryanhill1, line 33: **D1.** The guard is placed on cirq's outgoing edges, but cirq has three (`braket`, `qasm2`, `pytket`) and `pytket` is unguarded — so `transpile(parameterized_cirq, "braket")` still leaks `TypeError: cannot determine truth value of Relational...` from the `cirq→pytket→braket` hop, which is the issue's own reproduction.

Guarding `cirq→pytket` is the wrong fix: `transpile(parameterized_cirq, "pytket")` legitimately succeeds keeping the symbol (verified), like the `cirq→pyquil` path you left alone. The complete fix validates at the concrete-target ingest — add the same `free_symbols()` check to `pytket_to_braket` and `pytket_to_qasm2` — so any symbolic source is caught, not just cirq's direct edges.

ryanhill1 (changes requested):
Thanks @danielgaskins — I confirmed the bug on current `main` (cirq 1.6.1 leaks raw `TypeError`s, just different strings than 1.4.1), and the direct `cirq_to_braket` / `cirq_to_qasm2` fixes are correct and clean. The differential-testing context in the issue was genuinely useful too.

But the guard only covers 2 of cirq's 3 outgoing edges — `cirq → pytket` is unguarded — so `transpile()` still routes around it. Two problems, both verified with the full dev requirements (`pytket-cirq` is unconditional in `requirements-dev`, so CI installs it):

- **B1 (blocking):** `transpile(parameterized_cirq, "qasm2")` routes `cirq→pytket→qasm2` and **silently returns invalid QASM** (`rx((theta/pi)*pi) q[0];`, symbol as text) instead of raising — worse than the original bug. Your own `test_...[qasm2]` fails on it (3/3 deterministic), so CI will be red once workflows run.
- **D1:** `transpile(parameterized_cirq, "braket")` — the issue's own repro — still leaks a raw `TypeError: cannot determine truth value of Relational...` from the `cirq→pytket→braket` hop.

Fix: don't guard `cirq→pytket` (symbolic `cirq→pytket` is legitimate, like the `cirq→pyquil` path you correctly left alone — I verified it succeeds). Validate at the concrete-target *ingest* instead: `pytket_to_braket` and `pytket_to_qasm2` should reject symbolic input via `circuit.free_symbols()`, mirroring what you did for cirq.

tests/transpiler/cirq/test_conversions_cirq.py
danielgaskins: Updated the graph-level test to require an error for all three targets and to verify that no `TypeError` appears in the aggregated message. The PyTKET-to-QASM 2 boundary now rejects unresolved symbols, so the routed path can no longer emit symbolic QASM.

qbraid/transpiler/conversions/cirq/_utils.py
danielgaskins: Added shared PyTKET parameter validation at the Braket and OpenQASM 2 conversion boundaries. I left Cirq-to-PyTKET unchanged and added a regression confirming that it preserves unresolved symbols.

ryanhill1 (commented):
Thanks @danielgaskins — B1 and D1 are both fixed. Two more things to fix:

- **D1** `import sympy` sits above the `try:` in the pytket test module, so the
  file now errors at collection instead of skipping when pytket isn't installed.
- **D2** the guards also reject symbols that never reach the output — a symbolic
  global phase converted fine before and now raises.

**N1** is a one-liner in a file you already touched; **N2**–**N3** are take-or-leave
wording and tidiness.

tests/transpiler/pytket/test_conversions_pytket.py
```diff
@@ -18,6 +18,7 @@
 """
 
 import pytest
+import sympy
```
ryanhill1, line 21: **D1.** `sympy` is not a base dependency — it ships only with the `cirq` extra and
`requirements-dev` — so this import above the `try:` breaks the module's skip
contract. With pytket/cirq/sympy absent, the file raises
`ModuleNotFoundError: No module named 'sympy'` at collection instead of skipping;
`main` imports cleanly under the same conditions. CI won't catch it since dev
requirements always install sympy. Please move it inside the `try` block.

qbraid/transpiler/conversions/cirq/_utils.py
```diff
@@ -0,0 +1,38 @@
...
+    """Reject unresolved parameters before entering a concrete conversion.
+
+    Args:
+        circuit: Cirq circuit to validate.
+        target: Name of the concrete target format.
+
+    Raises:
+        ProgramConversionError: If the circuit contains unresolved parameters.
+    """
+    parameters = sorted(cirq.parameter_names(circuit))
```
ryanhill1, line 32: **D2.** `parameter_names()` counts symbols that never reach the output, so this
rejects a conversion that worked before:

```python
c = cirq.Circuit(cirq.H(q), cirq.global_phase_operation(sympy.exp(1j * t)))
cirq_to_qasm2(c)   # main: 'OPENQASM 2.0; ... h q[0];'   this PR: ProgramConversionError
```

Same on the pytket side via `free_symbols()` — `TKCircuit(1).H(0)` plus
`add_phase(theta)` converts fine through both `tk_to_braket` and
`circuit_to_qasm_str` today, and is now rejected. Global phase is unobservable and
both writers already drop it. Either exclude global-phase-only symbols here, or
decide it's acceptable and say so in the changelog entry — but it shouldn't land
silently.

qbraid/transpiler/conversions/pytket/pytket_extras.py
```diff
@@ -52,9 +54,13 @@ def pytket_to_braket(circuit: pytket.circuit.Circuit) -> braket.circuits.Circuit
 
     Returns:
         braket.circuits.Circuit: Braket circuit equivalent to input pytket circuit.
+
+    Raises:
+        ProgramConversionError: If the circuit contains unresolved parameters.
     """
     from pytket.circuit import OpType  # pylint: disable=import-outside-toplevel
 
+    _validate_resolved_parameters(circuit, "Amazon Braket")
```
ryanhill1, line 63: **N1.** `pytket_to_pyqir` in this same file is one line away from the helper and
still leaks the raw error for a symbolic circuit:

```
TypeError: argument 'theta': failed to extract enum Angle ('Value | Constant')
```

QIR is as concrete a target as Braket. Worth guarding while you're here.
`cirq_to_qat` (`WrongArgumentType: 3.14159265358979*theta/pi is not compatible
with type <class 'float'>`) and `cirq_to_pyqir` (`CirqConversionError: Failed to
preprocess circuit.`, which never mentions parameters) are the same gap on the
cirq side, plus `cirq_to_stim` — all pre-existing, so fine to leave for a
follow-up if you'd rather keep this PR tight.

qbraid/transpiler/conversions/pytket/pytket_extras.py
```diff
@@ -52,9 +54,13 @@ def pytket_to_braket(circuit: pytket.circuit.Circuit) -> braket.circuits.Circuit
 
     Returns:
         braket.circuits.Circuit: Braket circuit equivalent to input pytket circuit.
+
+    Raises:
```
ryanhill1, line 58: **N2.** This function already raises `ProgramConversionError` for two other
reasons — mid-circuit measurement, and two qubits measured into one bit — so a
`Raises:` naming only unresolved parameters reads as the complete list.
`cirq_to_braket` has five other raise sites with the same problem. Either widen
the wording or drop the section.

qbraid/transpiler/conversions/pytket/_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    """Reject unresolved parameters before entering a concrete conversion.
+
+    Args:
+        circuit: PyTKET circuit to validate.
+        target: Name of the concrete target format.
+
+    Raises:
+        ProgramConversionError: If the circuit contains unresolved parameters.
+    """
+    parameters = sorted(str(symbol) for symbol in circuit.free_symbols())
```
ryanhill1: **N3.** This and `cirq/_utils.py` are the same function under the same name,
differing only in the symbol accessor and the word Cirq/PyTKET. The message
construction could be one shared helper taking `(names, source, target)`, with
each module supplying its own accessor.

Also: this header says 2026, `cirq/_utils.py` says 2025. The check doesn't enforce
it, but 2026 matches the other files added this cycle.

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ Writing an entry:
...
 - Fixed OpenQASM programs being rejected before submission for AWS- and Open Quantum-hosted IonQ devices when they used constructs IonQ JSON cannot express, a Braket verbatim box above all. Both vendors forward the program as written, so the IonQ JSON check no longer applies to them ([#1386](https://github.com/qBraid/qBraid/pull/1386))
 - Fixed `Result.data.get_counts()` raising `TypeError` for Azure Pasqal jobs. Pasqal's emulators wrap the histogram as `{"counter": ..., "raw": ...}`, which was passed through instead of unwrapped, so no measurement counts were reachable ([#1381](https://github.com/qBraid/qBraid/pull/1381))
+- Cirq and PyTKET circuits with unresolved parameters now raise `ProgramConversionError` naming those parameters when converting to Braket or OpenQASM 2, instead of leaking NumPy or SymPy `TypeError`s ([#1363](https://github.com/qBraid/qBraid/issues/1363))
```
ryanhill1, line 62: **N3.** Link the PR instead of the issue at the end.

The wording is also slightly off for the new half: "instead of leaking NumPy or
SymPy `TypeError`s" describes the cirq edges, but `pytket -> qasm2` previously
emitted *invalid QASM* (`rx((theta)*pi) q[0];`) with no error at all. That
something which appeared to work now raises is the part a reader scanning for
"does this affect me?" needs.

tests/transpiler/pytket/test_conversions_pytket.py
danielgaskins: Sure thing, I moved `sympy` inside the guarded import block. I also changed the new parametrized tests to resolve converter names only when the test runs, so the module still collects and skips when the optional dependencies are absent.

qbraid/transpiler/conversions/cirq/_utils.py
danielgaskins: I excluded parameters used only by discarded global-phase operations. The validators now inspect gate operations, so a parameter that also appears on a gate is still rejected. I added regression coverage for the Cirq QASM 2 and PyTKET Braket and QASM 2 cases.

qbraid/transpiler/conversions/pytket/pytket_extras.py
danielgaskins: I added the same validation to `pytket_to_pyqir` and included it in the unresolved-parameter test. I left the pre-existing Cirq QAT, PyQIR, and Stim cases for a separate follow-up so this PR stays focused.

qbraid/transpiler/conversions/pytket/pytket_extras.py
danielgaskins: I widened both Braket `Raises` descriptions to cover any circuit that cannot be represented, rather than listing only unresolved parameters.

CHANGELOG.md
danielgaskins: Updated. The changelog now distinguishes Cirq's numeric errors from PyTKET-to-QASM 2 emitting invalid parameterized QASM, notes that global-phase-only symbols remain accepted, and links this PR.

ryanhill1 (approved):

PR #1384, opened 2026-09-08
feat(runtime): add resource estimate client (qbraid.runtime.estimate)

Kenny-Heitritter:
## Parent
qBraid/qbraid-simulators#240 — resource/cost estimator v1 (slice 8). Depends on qBraid/qbraid-core#316 (0.6.1) being released — `requirements.txt` pin bumped to `qbraid-core>=0.6.1,<1.0`.

## What
```python
est = qbraid.runtime.estimate(program, targets=None, shots=None, target=None, narrate=True, wait=True)
est.frontier / est.infeasible / est.open_questions / est.narrative / est.features
est.to_dataframe()   # one row per frontier point; cost split exact / runtime / total; weakest `basis` column
est.refresh(); est.wait(timeout=300, poll=2.0)   # raises on FAILED
```
Also `QbraidProvider.estimate(...)` and `QbraidEstimate`. Program conversion reuses the `device.run` path (QASM strings pass through verbatim; other programs transpile to qasm3). pandas imported lazily.

## Verification
`pytest tests -q` 2,245 passed / 654 skipped · pylint, isort, black, ruff, headers pass · mypy: 869 pre-existing baseline errors, zero added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Writing an entry:
 ## [Unreleased]
 
 ### Added
+- Added `qbraid.runtime.estimate()` and `QbraidProvider.estimate()` for resource and cost frontiers, with polling, open questions, narration, and optional pandas export through `QbraidEstimate`.
 - Added `QuantumDevice.supported_run_inputs` method, which returns the sorted list of program type aliases that can be passed as `run_input` to the device's `run` method ([#803](https://github.com/qBraid/qBraid/issues/803))
 - Added `QudoraProvider`, `QudoraDevice`, and `QudoraJob` classes implementing the qBraid runtime interface for the [QUDORA Cloud](https://api.qudora.com). Authenticates with a `Bearer` API token and submits OpenQASM directly over REST (no vendor SDK); reuses qBraid's existing `qasm2`/`qasm3` program specs so no converter is required. Device ids are backend `username`s (email addresses), and a program list is submitted as a single batched QUDORA job returning one histogram per program ([#1292](https://github.com/qBraid/qBraid/pull/1292))
 - Added `QbraidJob.compiled_program()`, returning the vendor-compiled program that executed on the QPU as a `Program` (`format` is `"qasm3"` for IQM, `"quil"` for Rigetti). It reveals the physical qubits selected and the logical-to-physical mapping the vendor's compiler chose. Returns `None` when no compiled program exists — a simulator, a job that has not completed, or one submitted before the backend captured them — so absence does not need a `try`/`except`. Requires `qbraid-core>=0.3.9`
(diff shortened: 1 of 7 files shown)


ryanhill1:
Two notes for when qBraid/qbraid-core#316 lands. No changes needed to this branch's code — I ran its tests against #316 as it currently stands (main merged in, plus the schema constraints from review) and got `24 passed, 1 skipped` here and `95 passed, 6 skipped` across `tests/runtime/test_estimate.py` + `tests/runtime/native`. The fixtures already satisfy the new bounds: features are all `>= 0`, and quality appears as `{"value": 1, ...}` and `{"value": None, "basis": "unavailable"}` — the first sits exactly on the `<= 1` boundary, the second is skipped.

## 1. The pin needs updating before this merges

`requirements.txt` moves to `qbraid-core>=0.6.1,<1.0`. Released core is **0.6.12**, which satisfies that and contains none of these schemas. Against it:

```
qbraid/runtime/native/__init__.py:47: in <module>
    from .estimate import QbraidEstimate
qbraid/runtime/native/estimate.py:24: in <module>
    from qbraid_core.services.runtime.schemas.estimate import (
E   ModuleNotFoundError: No module named 'qbraid_core.services.runtime.schemas.estimate'
```

That import is in `qbraid/runtime/native/__init__.py`, so the whole `qbraid.runtime.native` package fails to import — not just the estimate feature. Merging with this pin would break the SDK for anyone on a current qbraid-core.

#316 no longer carries a version bump (it was `0.6.1`, stale against main's `0.6.12`, so it was dropped in favour of main's). Version bumps happen in their own PR here, so the order is: merge #316 → bump to 0.6.13 → release → then set this pin to `>=0.6.13,<1.0`. The exact floor isn't knowable until the bump lands.

The PR description also still reads "Depends on qBraid/qbraid-core#316 (0.6.1) being released" — worth correcting alongside the pin, since that's the sentence the wrong floor came from.

## 2. Validation now fires at the SDK's parse boundary

`provider.py:316` does `Estimate.model_validate(response.json()["data"])`, and #316 adds `ge=0` on the `CircuitFeatures` counts plus a `[0, 1]` check on `FrontierPoint.quality`.

Worth knowing because the estimator node can produce a quality outside that range: `engine/schemas.Calibration.meanTwoQubitGateError` and `meanReadoutError` are unbounded floats (qBraid/qbraid-simulators#248), and quality is `(1 - e2) ** twoQubitGateCount * (1 - er) ** numQubits`. A negative error rate gives quality above 1; an error above 1 gives a negative quality.

If that reaches the API, this client raises rather than surfacing an odd number — and since `frontier` is a list, one bad device makes the **whole** `Estimate` unparseable instead of one row looking strange. That's the right trade for a contract with no deployed consumers yet, but the durable fix is bounding the calibration inputs at the node before this ships to users.

Same shape as qBraid/qBraid#1425, where the unbounded fields make `best_qubits()` rank a corrupt qubit as the best on the device.

PR #1391, opened 2026-09-09
fix(qudora): raise a job error, not a TypeError, on a null histogram

ryanhill1:
### What this does

`result()` guards a missing payload with `if not result_payload`, which `[None]` passes — a non-empty list is truthy. The null then reaches `json.loads` in `_parse_counts`, so the caller gets

```
TypeError: the JSON object must be str, bytes or bytearray, not NoneType
```

instead of an error naming the job and what was wrong with it. This widens the existing guard to reject null entries too, raising the `QudoraJobError` it already raises for an absent payload.

### This is hardening, not a bug fix — and the original premise was wrong

This PR started out much larger, on the premise (taken from a comment in qBraid/qbraid-runtime-api#257) that QUDORA flips a job to `Completed` shortly *before* writing its result rows, leaving a `[null]` placeholder in between. I tested that rather than assume it, polling live jobs every 50ms across both published devices — 8 jobs, 5 programs each, ~450 samples:

```
   282x  status=Submitted    result=all-null(len 5)
    83x  status=Running      result=PARTIAL 1/5 .. 4/5
    30x  status=Running      result=all-null(len 5)
     2x  status=Running      result=written(len 5)
     8x  status=Completed    result=written(len 5)

Completed-with-nulls observed: 0
```

The rows are filled **while the job is Running**, and the flip to `Completed` follows — the `Running` + fully-written samples show that ordering directly. Every first-`Completed` read already carried its full payload. So `include_results=True` on a completed job does return the results, and the race does not appear to exist.

What remains is a decode-path guard: `[None]` is a shape the API demonstrably produces (a cancelled job's record carries it), and if it ever reached `result()` the failure would be a `TypeError` from the JSON layer rather than something a caller can act on. Two lines, no waiting, no timeout, no behaviour change for any observed state.

No changelog entry: nothing user-visible changes for the states QUDORA actually produces.

Close this if you would rather not carry a guard for a state that has not been seen — that is a consistent position and the measurements above are the useful part either way.

### DIFF
--- a/qbraid/runtime/qudora/job.py
+++ b/qbraid/runtime/qudora/job.py
@@ -133,13 +133,20 @@ class QudoraJob(QuantumJob):
             message = job_data["user_error"] or f"job ended with status {status.name}"
             raise QudoraJobError(f"QUDORA job {self.id} did not complete: {message}.")
 
-        # A null entry means that program has no histogram to decode. Caught here so it
-        # surfaces as this function's own error rather than a TypeError out of
-        # ``json.loads`` -- ``[None]`` slips past a bare falsiness check, a non-empty
-        # list being truthy.
         result_payload = job_data["result"]
-        if not result_payload or any(entry is None for entry in result_payload):
+        if not result_payload:
             raise QudoraJobError(f"QUDORA job {self.id} completed but returned no result data.")
+
+        # A null entry is one program with no histogram to decode, which a bare falsiness
+        # check misses -- a list with a null in it is still truthy. Caught here so it
+        # surfaces as this function's own error, naming how much of the batch is missing,
+        # rather than a TypeError out of ``json.loads``.
+        missing = sum(1 for entry in result_payload if entry is None)
+        if missing:
+            raise QudoraJobError(
+                f"QUDORA job {self.id} completed but returned incomplete result data: "
+                f"{missing} of {len(result_payload)} programs have no histogram."
+            )
 
         data = GateModelResultData(measurement_counts=self._parse_counts(result_payload))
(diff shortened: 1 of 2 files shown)


qbraid/runtime/qudora/job.py
```diff
@@ -133,8 +133,12 @@ def result(self) -> Result:
...
             raise QudoraJobError(f"QUDORA job {self.id} did not complete: {message}.")
 
+        # A null entry means that program has no histogram to decode. Caught here so it
+        # surfaces as this function's own error rather than a TypeError out of
+        # ``json.loads`` -- ``[None]`` slips past a bare falsiness check, a non-empty
+        # list being truthy.
         result_payload = job_data["result"]
-        if not result_payload:
+        if not result_payload or any(entry is None for entry in result_payload):
             raise QudoraJobError(f"QUDORA job {self.id} completed but returned no result data.")
```
ryanhill1: should it instead say 'incomplete' job data for the latter?

ryanhill1: Good catch — done in ac78533d. An incomplete payload and an absent one are different failures, and collapsing them lost the distinction.

Split into two guards. Empty or missing keeps the original wording; a payload with nulls now names how much of the batch is missing:

```
QUDORA job 42 completed but returned incomplete result data: 1 of 5 programs have no histogram.
```

So a single failed program in a five-program batch reads as that rather than as a total loss, which also tells you whether re-running the whole batch is warranted. Test is parametrized over the single, partial-pair and partial-batch shapes to pin the counts.

PR #1405, opened 2026-09-18
Fix qubit numbering for multiple registers in qasm -> IonQ conversion

### LINKED ISSUE #770: Add support for multiple `qreg` definitions in `qasm2_to_ionq`
### Environment

- **qBraid version**: 0.8.2
- **Python version**: 3.12
- **Operating system**: macOS 14.4

### What happened?

For the following program - 

```python3 
from qbraid.transpiler.conversions.qasm2 import qasm2_to_ionq

qasm_str = """
OPENQASM 2.0;
include "qelib1.inc";
qreg q[2];
x q[0];
y q[0];

qreg q2[2];
cx q2[0], q2[1];

"""

print(qasm2_to_ionq(qasm_str))
```

the output is - 

```python3
{'qubits': 4, 'circuit': [{'gate': 'x', 'target': 0}, 
                        {'gate': 'y', 'target': 0},
                        {'gate': 'cnot', 'control': 0, 'target': 1}]}
```

The expected output for the program should be - 

```python3
{'qubits': 4, 'circuit': [{'gate': 'x', 'target': 0},
                        {'gate': 'y', 'target': 0}, 
                        {'gate': 'cnot', 'control': 2, 'target': 3}]}
```
as `cnot` is applied on the 2nd and 3rd physical qubit (`q2[0]` and `q2[1]`)

### Suggestions (Optional)

There needs to be correct memory of the `qreg`s being passed in the qasm2 program and mapping should be applied on the qubits when they are used in the converter. 

Possibly this can benefit from the [semantic analyser of qbraid-qir](https://github.com/qBraid/qbraid-qir/issues/148#issuecomment-2382361643) and can be taken up when it is available.

nero-harsh:
## Summary of changes

Closes #770.

`openqasm3_to_ionq` (which `qasm2_to_ionq` delegates to) took each qubit's index within its own register and emitted it unchanged, so every register's qubits were numbered from 0. With more than one register the output addressed the wrong qubits:

```python
qreg q[2];
qreg q2[2];
cx q2[0], q2[1];   # emitted as cnot control 0 target 1, should be 2 -> 3
h q2;              # emitted on 0, 1, should be 2, 3
```

Across registers it can emit a gate with control == target. The Deutsch-Jozsa fixture in `test_qasm_to_ionq.py` shows this: `qubit[4] q; qubit ancilla;` produced `cnot control 0 target 0` for the ancilla, and the test asserted it.

The fix adds `_register_offsets`, which walks the `QubitDeclaration`s of the AST being converted and records where each register starts, in declaration order (the same order pyqasm uses for `num_qubits`). Both the indexed-qubit branch and the bare-register branch add that offset. Sizes come from `_qubit_registers` when pyqasm has populated it, so const-sized registers like `qubit[n] b` work, and fall back to the literal in the declaration when it hasn't. Existing tests call `_parse_gates` on programs pyqasm rejects (e.g. `ms` with 4 params), and those still get the intended `ValueError` rather than a `KeyError`.

Tests:
- `test_qasm2_to_ionq_multiple_registers`: the program from the issue plus a bare-register gate and a cross-register cx.
- `test_qasm3_to_ionq_multiple_registers`: single-qubit, const-sized, and literal-sized registers.
- `deutch_jozsa_ionq` fixture updated so the ancilla is qubit 4.

The two new tests and the three DJ tests fail on `main` and pass with the change. The full suite (`--remote false`, core optional deps installed) has the same failures/errors with and without this change, all from optional packages I don't have installed, plus 2 more passing. black (<=25.12), isort and pylint are clean on both files.


qbraid/transpiler/conversions/openqasm3/openqasm3_to_ionq.py
```diff
@@ -148,6 +148,31 @@ def _parse_angle(angle: str, gate_name: str) -> float:
...
+        if not isinstance(statement, openqasm3.ast.QubitDeclaration):
+            continue
+        name = statement.qubit.name
+        offsets[name] = offset
+        if name in register_sizes:
+            offset += register_sizes[name]
+        elif statement.size is None:
+            offset += 1
+        else:
+            offset += statement.size.value
```
ryanhill1, line 172: **D1a** — when `unroll()` fails *before* this declaration is processed, the name
never reaches `_qubit_registers`, so we fall through to here — and a const- or
expression-sized register has an `Identifier`, not an `IntegerLiteral`:

```qasm
qubit[2] a;
ms(0.1, 0.2, 0.3, 0.4) a[0], a[1];   // unroll fails here
const int n = 2;
qubit[n] b;
h b[0];
```

`main` reports `ValueError: Invalid number of parameters for the 'ms' gate...`;
with this branch it's `AttributeError: 'Identifier' object has no attribute 'value'`.
Because `_register_offsets` runs eagerly at the top of `_parse_gates`, it throws
before the gate loop reaches the real error.

A `getattr(statement.size, "value", None)` with a skip when it's absent would do it.
Worth a test too — I checked by making this line raise, and the three tests that
reach this branch all use literal sizes, so the non-literal path is uncovered.

qbraid/transpiler/conversions/openqasm3/openqasm3_to_ionq.py
```diff
@@ -204,13 +231,15 @@ def _parse_gates(program: Union[OpenQasm2Program, OpenQasm3Program]) -> list[dic
                 reg_name = qubits[0].name
                 for qreg_name, reg_size in program_qubits:
                     if qreg_name == reg_name:
-                        qubit_values = list(range(reg_size))
+                        start = register_offsets[qreg_name]
+                        qubit_values = list(range(start, start + reg_size))
                         break
             else:
                 for qubit in qubits:
+                    start = register_offsets[qubit.name.name]
```
ryanhill1, line 239: **D1b** — this `KeyError`s on any operand whose name isn't a top-level
`QubitDeclaration`. An OpenQASM 3 alias is the clearest case:

```qasm
qubit[4] q;
let a = q[0:1];
cx a[0], a[1];     // KeyError: 'a'   (main: pyqasm ValidationError)
```

`qubit.name.name` also assumes an `IndexedIdentifier`; for a bare `Identifier`
(`cx q, r`) it's `'str' object has no attribute 'name'`. That one fails on main too,
so it isn't a regression — but a `.get()` with a 0 fallback, or an explicit check
that resolves unknown names before indexing, would cover both this and D1a.

ryanhill1 (changes requested):
Thanks for this.

I verified the numbering convention independently rather than against the issue's
expected values: built the same two-register circuit in qiskit and compared
`find_bit` global indices, and declaration-order concatenation matches exactly.
Also confirmed the fix is complete for #770 (the two index sites you changed are
the only ones in the file; `qubits` already came from `num_qubits`), that no other
test in the repo asserts IonQ output for a multi-register program, and that the
`deutch_jozsa_ionq` change is a genuine bug fix rather than a test edited to fit —
the old expectation asserted `{"gate": "cnot", "control": 0, "target": 0}`, a
self-targeting CNOT. black, isort and pylint are clean; 46 pass here, 6 fail on main.

One thing to fix — **D1**, with two inline comments below.

### D1 — `_register_offsets` assumes more about operands and sizes than holds

Two assumptions: every gate operand's name resolves to a top-level
`QubitDeclaration`, and every declaration's `size` is an `IntegerLiteral`. Three
programs break one or the other, and each turns a typed error into a raw one:

```qasm
// (a) non-literal size, declared after the point where unroll() fails
qubit[2] a;
ms(0.1, 0.2, 0.3, 0.4) a[0], a[1];
const int n = 2;
qubit[n] b;
h b[0];
// main: ValueError: Invalid number of parameters for the 'ms' gate...
// here: AttributeError: 'Identifier' object has no attribute 'value'
```

```qasm
// (b) alias — 'a' is a `let`, never a QubitDeclaration
qubit[4] q;
let a = q[0:1];
cx a[0], a[1];
// main: ValidationError: Index 1 out of range...
// here: KeyError: 'a'
```

(c) is `cx q, r` (bare-register broadcast), which fails both before and after —
only the message changes — but it shares the same root cause, so a fix for (a)
and (b) should cover it.

Worth flagging against the PR description: (b) is the `KeyError` the description
says is avoided. The `ms` case you checked goes down a different path.

**Severity:** nothing that converted successfully before regresses. Single-register
programs get offset 0 and are byte-identical, and every case above already failed
on main — it's the error *type* that degrades, not the result.

nero-harsh:
Thanks for the thorough check, especially verifying the offset convention independently against qiskit rather than trusting the issue's numbers.

Fixed D1a and D1b, both by failing loudly with a typed `ValueError` instead of the raw `AttributeError`/`KeyError`:

- **D1a**: `_register_offsets` now checks `isinstance(statement.size, openqasm3.ast.IntegerLiteral)` before reading `.value`. A non-literal size that pyqasm hasn't resolved raises `ValueError` naming the register.
- **D1b**: the multi-qubit branch now checks the operand is an `IndexedIdentifier` before reading `.name.name`, and checks the resolved name is actually in `register_offsets` before indexing it. An alias or a bare whole-register operand (your case c, same root cause as you said) each raise a `ValueError` naming the problem instead of crashing internally.

Corrected the PR description too: the alias case is the one that raises what used to be a `KeyError`, not case (a).

Added three regression tests reproducing (a), (b) and (c) from your comment, confirmed they fail on the pre-fix code and pass now, plus the case-c test also checks that the public `qasm3_to_ionq` wrapper still recovers via its pyqasm-assisted retry. Rebased onto current `main` to clear the changelog conflict; 49 pass locally (46 existing + 3 new), black/isort/pylint clean.

qbraid/transpiler/conversions/openqasm3/openqasm3_to_ionq.py
```diff
@@ -204,13 +239,30 @@ def _parse_gates(program: Union[OpenQasm2Program, OpenQasm3Program]) -> list[dic
                 reg_name = qubits[0].name
                 for qreg_name, reg_size in program_qubits:
                     if qreg_name == reg_name:
-                        qubit_values = list(range(reg_size))
+                        start = register_offsets[qreg_name]
+                        qubit_values = list(range(start, start + reg_size))
                         break
             else:
```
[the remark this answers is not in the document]
ryanhill1: Good catch, and the bug is real — I reproduced it:

```qasm
qubit[4] q;
let a = q[0:1];
h a;
x q[3];
```
```
circuit=[{"gate": "x", "target": 3}]     # the h is gone, no error
```

Tracking it separately as #1421, because it is not this PR: `origin/main` produces byte-identical output for that program, and the only change this PR makes inside that branch is the offset arithmetic it exists to fix —

```diff
-                        qubit_values = list(range(reg_size))
+                        start = register_offsets[qreg_name]
+                        qubit_values = list(range(start, start + reg_size))
```

The missing `else` on the name search is untouched. The new explicit checks in the adjacent multi-qubit branch are what drew attention to it.

One note on the suggested patch: raising for any unresolved bare register would also reject an alias that arguably ought to convert. Converting the alias on its own yields `[{"gate": "h", "target": 0}]` — one gate where `a` spans two qubits — so there looks to be a second defect in the same code. Both are captured in #1421 so they can be fixed together rather than guarded here.

ryanhill1 (approved):

PR #1411, opened 2026-09-19
Fix multi-statement if blocks losing their condition in QASM -> Cirq

nero-harsh:
## Summary of changes

`normalize_if_blocks` (used by `qasm3_to_cirq` and `qasm2_to_cirq` to turn pyqasm's braced if blocks into Cirq-parsable QASM 2 `if()` lines) treats the whole block body as one statement. For a block with more than one statement, only the first stays conditional:

```
if (c[0] == false) {
  x q[1];
  x q[2];
}
```
becomes
```
if(c==0) x q[1];
  x q[2];
```

End to end through `qasm3_to_cirq`, the `x q[2]` comes out as a plain unconditional `X`, with no error. The circuit silently does something different from the input program, which is what the converter contract in `docs/transpiler.md` rules out.

The fix splits the body on `;` and emits one `if(reg==val) stmt;` per statement. QASM 2 has no block form, so repeating the condition is the direct translation.

One limitation I didn't change: if a block measured into its own condition register and then used it (`if (c == 1) { c[0] = measure q[0]; x q[1]; }`), re-checking the condition per line would read the new value. The old code had the same problem and worse, and I didn't find pyqasm emitting that shape. Happy to make that case raise instead if you'd prefer.

Tests:
- `test_normalize_if_blocks_multiple_statements` in `tests/passes/test_qasm3_compat.py`
- `test_qasm3_to_cirq_multi_statement_if_block` in `tests/transpiler/cirq/test_from_qasm.py`, which checks that both X gates come out as `ClassicallyControlledOperation`s

Both fail on `main` and pass with the change. The existing `normalize_if_blocks` and Cirq conditional tests are unchanged and pass. `tests/passes` and `tests/transpiler` give the same failure/error counts with and without the change, all from optional packages I don't have installed, plus the 2 new passing tests. black, isort and pylint are clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed QASM 2 and QASM 3 conversions to Cirq so every statement within a multi-statement conditional block is correctly controlled by the condition.
  - Ensured gates following the first statement in an `if` block no longer execute unconditionally.
  - Added regression coverage for multi-statement conditional blocks and preserved expected behavior for measurements and unconditional gates.


ryanhill1 (approved):

